PR #34, opened 2017-06-21
Add semantic model

Pzixel:
Today, we are not able to access semantic model which is required sometimes. Assume we change your example like following:

```csharp
[DuplicateWithSuffix("A")]
public class Foo
{
  [AnotherUsefulAttribute("B")]
  public int Foo() => 0;
}
```
With current architecture we are unable to get attributes value in normal way, because we require semantic model to convert `AttributeSynax` to `AttributeData`. This merge request contains all required changes and proof-of-concept that it works as expected


AArnott:
Thanks.

Issue #26, opened 2017-06-02
Unable to build sample project in Visual Studio


Pzixel:
I always get the same error:

```
6>------ Rebuild All started: Project: CodeGeneration.Roslyn.Tasks, Configuration: Debug Any CPU ------
6>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(4825,5): warning MSB3061: Unable to delete file "C:\Users\Alex\Source\Repos\CodeGeneration.Roslyn\bin\CodeGeneration.Roslyn.Tasks\Debug\net46\CodeGeneration.Roslyn.Tasks.dll". Access to the path 'C:\Users\Alex\Source\Repos\CodeGeneration.Roslyn\bin\CodeGeneration.Roslyn.Tasks\Debug\net46\CodeGeneration.Roslyn.Tasks.dll' is denied.
6>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(4148,5): warning MSB3026: Could not copy "C:\Users\Alex\Source\Repos\CodeGeneration.Roslyn\bin\CodeGeneration.Roslyn.Tasks.Helper\Debug\net46\CodeGeneration.Roslyn.Tasks.Helper.dll" to "C:\Users\Alex\Source\Repos\CodeGeneration.Roslyn\src\..\bin\CodeGeneration.Roslyn.Tasks\Debug\net46\CodeGeneration.Roslyn.Tasks.Helper.dll". Beginning retry 1 in 1000ms. The process cannot access the file 'C:\Users\Alex\Source\Repos\CodeGeneration.Roslyn\src\..\bin\CodeGeneration.Roslyn.Tasks\Debug\net46\CodeGeneration.Roslyn.Tasks.Helper.dll' because it is being used by another process.
[...]

PR #39, opened 2017-07-04
Nuget update Microsoft.CodeAnalysis.CSharp to 2.2.0

Pzixel:
I found that I'm unable to use C# 7.0 features such as `throw expression` or value tuples. If I update packages locally I'm unable to build solution due to fails in nuget dependency consolidation mechanism. So I propose to update your nuget up to date. It doesn't require any additional dependencies, doesn't require higher NetStandard, so I don't see any problems here.

### DIFF
--- a/src/CodeGeneration.Roslyn/CodeGeneration.Roslyn.csproj
+++ b/src/CodeGeneration.Roslyn/CodeGeneration.Roslyn.csproj
@@ -16,7 +16,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="1.3.2" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.2.0" />
     <PackageReference Include="Validation" Version="2.4.13" />
   </ItemGroup>
 


AArnott:
Thanks for contributing

Pzixel:
I glad to contribute :) Waiting for nuget package.

AArnott:
published

Pzixel:
That's weird but my problem still here. After updating to 2.2.0 it doesn't work. See pull request, the only change is updating to 0.4.11: https://github.com/Pzixel/RemoteClient.Roslyn/pull/3/files . I also changed NetStandard to 1.5, but it doesn't matter, it doesn't work with 1.6 too.
And here is build of master which is ok: https://github.com/Pzixel/RemoteClient.Roslyn/tree/master

I investigated for a while but found nothing. Moving to 2.2.0 is breaking code somehow. Even if my generator do nothing except returning an empty class declaration without any complex logic.

AArnott:
OK. I thought you'd already tested this and found that it resolved your issue. For now, I've unlisted the 0.4.11 release from nuget.org to avoid it breaking anyone since I didn't personally test it either. 
It's still there on nuget.org so you can keep trying if you want by naming the version exactly..

Pzixel:
@AArnott I have checked it for some simple tests like included in repository (DuplicateWIthSuffix and so on). I was expected that it fails due to failure of dependency resolve mechanism. Sorry if I break your flow somehow. All checks passed so I was expected that it works fine. I didn't check it (as I should, sorry) on very mine code. 

I'l continue investigating this issue tomorrow. Sorry again for breaking things. But it's because I can build it locally (see #26 )

Pzixel:
That's bizzare. All samples are working fine in CodeGeneration.Roslyn solution: as you can see, all checks have passed. However, when it's included as a nuget package, I get this error even for very first example with `DuplicateWithSuffixAttribute`. Can't figure out what's wrong and why it's not working as nuget but working fine as project reference.

Pzixel:
After some investigation I found that there is weird problem with nuget:

image

For unknown reason it says that we are referencing both 1.3.1 and 2.2.0. I didn't find any reference to 1.3.1 in solution or its dependency...

Pzixel:
@AArnott You can publish this package back. It's my fault: dotnet-codegen 0.4.9 is not compatible with CodeAnalysis 2.2.0. `<DotNetCliToolReference Include="dotnet-codegen" Version="0.4.11" />` solves the problem.

AArnott:
Great. Thanks for the update. I've re-listed 0.4.11 packages.

Issue #44, opened 2017-07-12
Reporting progress and dealing with internal errors


dpvreony:
Hi

First off thank you for your efforts on this tool.

I was having an issue where my generator was throwing an exception (my fault :)) and there was a lack of detail coming from Visual Studio.

![20170712-dotnet-error](https://user-images.githubusercontent.com/170983/28106654-9dbc19c4-66dc-11e7-9591-035e17fcc668.PNG)

I've prepared a branch with the following changes

- Progress object is no longer null and reporting of diagnostic to stdout.
- try\catch around each file
- addition of a basic diagnostic that reports the exception during transform

This sees the following detail added to Visual Studio

![20170712-vserrors](https://user-images.githubusercontent.com/170983/28106758-011a1dfe-66dd-11e7-8da8-f7c20486fb7a.PNG)

PR #45, opened 2017-07-12
Addition of progress and per file error reporting

dpvreony:
As discussed in #44 

- Progress object is no longer null and reporting of diagnostic to stdout.
- try\catch around each file
- addition of a basic diagnostic that reports the exception during transform

Thanks


AArnott:
Looks great. Thanks for contributing, @dpvreony !

Issue #61, opened 2018-03-30
My generator class cannot find one of its dependencys


LokiMidgard:
I updated my Project from version `0.3.9` of this library to current `0.4.11`. The Sample `DuplicateWithSuffix` is working, but my generator stopped working.

It does not produce any kind of error output. After some debugging I think the problem is that the Type of my Generator cannot be loaded because a dependency is missing.

When I create an instance of my Generator in the constructor of `DuplicateWithSuffixGenerator` the build process prints an exception stack trace. Which claims the assembly `NDProperty.Core` can't be found. This is another project my Generator depends on.

```
1>Target GenerateCodeFromAttributes:
1>  Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.IO.FileNotFoundException: Could not load file or assembly 'NDProperty.Core, Version=0.14.0.0, Culture=neutral, PublicKeyToken=null'. Das System kann die angegebene Datei nicht finden.
1>     at NDProperty.Generator.DuplicateWithSuffixGenerator..ctor(AttributeData attributeData)
1>     --- End of inner exception stack trace ---
1>     at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
1>     at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
1>     at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes, StackCrawlMark& stackMark)
1>     at System.Activator.CreateInstance(Type type, BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes)
1>     at System.Activator.CreateInstance(Type type, Object[] args)
1>     at CodeGeneration.Roslyn.DocumentTransform.<FindCodeGenerators>d__2.MoveNext()
1>     at CodeGeneration.Roslyn.DocumentTransform.<TransformAsync>d__1.MoveNext()
1>  --- End of stack trace from previous location where exception was thrown ---
1>     at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
1>     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #62, opened 2018-03-31
Added Assembly Resolve

LokiMidgard:
Added logic to resolve referenced assemblies of generator assembly.

This should prevent problems with generators that uses other assemblies described in #61.

To use `DependencyContext` it was necessary to push target framework version to netstandard 1.6.

### DIFF
--- a/src/CodeGeneration.Roslyn/CompilationGenerator.cs
+++ b/src/CodeGeneration.Roslyn/CompilationGenerator.cs
@@ -7,6 +7,8 @@ namespace CodeGeneration.Roslyn
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Text;
+    using Microsoft.Extensions.DependencyModel;
+    using Microsoft.Extensions.DependencyModel.Resolution;
     using System;
     using System.Collections.Generic;
     using System.Globalization;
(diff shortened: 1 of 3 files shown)


src/CodeGeneration.Roslyn/CompilationGenerator.cs
```diff
@@ -157,7 +159,40 @@ public void Generate(IProgress<Diagnostic> progress = null, CancellationToken ca
...
+            var dependencyContext = DependencyContext.Load(assembly);
+
+            var assemblyResolver = new CompositeCompilationAssemblyResolver(new ICompilationAssemblyResolver[]
+            {
+                new AppBaseCompilationAssemblyResolver(Path.GetDirectoryName(path)),
+                new ReferenceAssemblyPathResolver(),
+                new PackageCompilationAssemblyResolver()
+            });
+
+            loadContext.Resolving += resolveAssembly;
```
AArnott: I don't think we should be _adding_ a new resolver every time we load an assembly. Shouldn't we define and add the resolver just once at the start of the tool? Otherwise the scaling will be awful: each assembly load will add yet another resolver and after 5 assembly loads we'll have 5 resolvers all doing equivalent things running.

LokiMidgard: I think we need an resolver for each folder where the assembly could be. `CompositeCompilationAssemblyResolve` does not have a method to add additional resolver after construction. So I think there were two possiblitys.

1. We could use a Hashset to keep track of the direcotrys we have already an resolver and create a new one if needed.
2. We are able to get all pathes up front and only create one `CompositeCompilationAssemblyResolver` that has a resolver for every directory.

I'm not sure if number two is feasible, so I would stick to the first approach. 

Any Thoughts on this?

LokiMidgard: I tried a third way.

Now there will actually be one callback that resolves the assembly. the used `CompositeCompilationAssemblyResolve` will actually a chain of `CompositeCompilationAssemblyResolve`s each one referencing the previous one and the directory resolver. And the new `DependencyContext` will be merged with the old.

In addition we keep track of all paths that are added to the resolver, not adding the same path twice.

Also we keep a dictionary of all loaded assembly's, if the same assembly would be loaded a second time we just return the already loaded assembly.

LokiMidgard:
I also updated I also updated NedbankGitVersioning to `2.1.23` and removed the hack that was previous needed since Appveyor updated there build tools.

I hope you do not mind that I mix this up with this issue.

Issue #54, opened 2017-10-28
Doesn't work with Microsoft.NET.SDK based project


smaillet:
# Repro steps
1) Load TestRoslynGenerator solution [TestRoslynGenerator.zip](https://github.com/AArnott/CodeGeneration.Roslyn/files/1424633/TestRoslynGenerator.zip) into VS2017
   1) I've tried with 15.4.1  and 15.5.0 Preview 2 same results
2) Build the solution (F6)

## Expected
successful build

## Actual
Error	MSB6006	"dotnet" exited with code 1

### Additional Details
After enabling full diagnostics in MSBuild I found the call to dotnet.exe and a message that provides some further hint on the problem:
```
dotnet codegen 
-r C:\Users\USERNAMEXYZ\.nuget\packages\codegeneration.roslyn.attributes\0.4.11\lib\netstandard1.0\CodeGeneration.Roslyn.Attributes.dll 
-r C:\Users\USERNAMEXYZ\.nuget\packages\codegeneration.roslyn\0.4.11\lib\netstandard1.5\CodeGeneration.Roslyn.dll
[ Standard .NET lib references not shown for brevity]
-r "C:\Users\USERNAMEXYZ\Documents\Visual Studio 2017\Projects\TestRoslynGenerator\TestRoslynGenerator\bin\Debug\netstandard1.5\TestRoslynGenerator.dll" 
-r C:\Users\USERNAMEXYZ\.nuget\packages\validation\2.4.13\lib\netstandard1.3\Validation.dll 
--out obj\Debug\netstandard1.5\ 
--generatedFilesList obj\Debug\netstandard1.5\y1sshcfo.ivy 
-- Consumer.cs

No executable found matching command "dotnet-codegen"
```
I don't pretend to understand the dotnet.exe tooling well enough to know what that means. 

### Suggestion
Provide complete standalone sample solution(s) that are built and validated as part of CI builds so it is easier for users to try and can catch issues in the builds

StephenStrickland:
Currently blocked by this.

amis92:
@smaillet @StephenStrickland there are two changes you need to add:

in the folder with solution file you need to add `nugget.config` with contents as follows (specifically important is the NuGet feed for CoreFx experimental repo):
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
[...]

PR #67, opened 2018-04-19
Assembly attributes support

### LINKED ISSUE #66: Generate C# code based on non-C# source [possible PR]
I'm looking toward some way to generate some C# code based on my text file. I tried several approaches but they didn't work.

Is it possible in current infrastructure? Or maybe you have some vision of how it could be done? I can implement it and send a PR because I need it myself, but I'd like first to talk a bit about what's your opinion on this matter.

Pzixel:
This is an attempt to provide generation for a global attribute which have no type to sit on. 

It also adds `ProjectDirectory` attribute because if attributes have no type to interact it should still have some way to get information about project. This guy should fix #66 because with knowlenge of project directory I can recursively get all needed files and transform them as I need. Test example just creates a const public field with path to the `csproj`, however, actual code generator may use it to list all `.myextension` files and use them to create an output.

### DIFF
--- a/src/CodeGeneration.Roslyn.Tasks/GenerateCodeFromAttributes.cs
+++ b/src/CodeGeneration.Roslyn.Tasks/GenerateCodeFromAttributes.cs
@@ -23,6 +23,9 @@ namespace CodeGeneration.Roslyn.Tasks
         public ITaskItem[] GeneratorAssemblySearchPaths { get; set; }
 
         [Required]
+        public string ProjectDirectory { get; set; }
+
+        [Required]
         public string IntermediateOutputDirectory { get; set; }
 
         public string ToolLocationOverride { get; set; }
@@ -79,6 +82,9 @@ namespace CodeGeneration.Roslyn.Tasks
 
             argBuilder.AppendLine("--out");
             argBuilder.AppendLine(this.IntermediateOutputDirectory);
+
+            argBuilder.AppendLine("--projectDir");
+            argBuilder.AppendLine(ProjectDirectory);
 
             this.generatedCompileItemsFilePath = Path.Combine(this.IntermediateOutputDirectory, Path.GetRandomFileName());
             argBuilder.AppendLine("--generatedFilesList");
--- a/src/CodeGeneration.Roslyn.Tasks/GenerateCodeFromAttributes.cs
+++ b/src/CodeGeneration.Roslyn.Tasks/GenerateCodeFromAttributes.cs
@@ -84,7 +84,7 @@ namespace CodeGeneration.Roslyn.Tasks
             argBuilder.AppendLine(this.IntermediateOutputDirectory);
 
             argBuilder.AppendLine("--projectDir");
-            argBuilder.AppendLine(ProjectDirectory);
+            argBuilder.AppendLine(this.ProjectDirectory);
 
             this.generatedCompileItemsFilePath = Path.Combine(this.IntermediateOutputDirectory, Path.GetRandomFileName());
             argBuilder.AppendLine("--generatedFilesList");
--- a/src/CodeGeneration.Roslyn/CompilationGenerator.cs
+++ b/src/CodeGeneration.Roslyn/CompilationGenerator.cs
@@ -65,6 +65,8 @@ namespace CodeGeneration.Roslyn
         /// </summary>
         public IEnumerable<string> EmptyGeneratedFiles => this.emptyGeneratedFiles;
 
+        public string ProjectDirectory { get; set; }
+
         public void Generate(IProgress<Diagnostic> progress = null, CancellationToken cancellationToken = default(CancellationToken))
         {
             Verify.Operation(this.Compile != null, $"{nameof(Compile)} must be set first.");
@@ -104,6 +106,7 @@ namespace CodeGeneration.Roslyn
                                 var generatedSyntaxTree = DocumentTransform.TransformAsync(
                                     compilation,
                                     inputSyntaxTree,
+                                    ProjectDirectory,
                                     this.LoadAssembly,
                                     progress).GetAwaiter().GetResult();
 
--- a/src/CodeGeneration.Roslyn/CompilationGenerator.cs
+++ b/src/CodeGeneration.Roslyn/CompilationGenerator.cs
@@ -106,7 +106,7 @@ namespace CodeGeneration.Roslyn
                                 var generatedSyntaxTree = DocumentTransform.TransformAsync(
                                     compilation,
                                     inputSyntaxTree,
-                                    ProjectDirectory,
+                                    this.ProjectDirectory,
                                     this.LoadAssembly,
                                     progress).GetAwaiter().GetResult();
 
--- a/src/CodeGeneration.Roslyn/DocumentTransform.cs
+++ b/src/CodeGeneration.Roslyn/DocumentTransform.cs
@@ -114,18 +114,12 @@ namespace CodeGeneration.Roslyn
             return compilationUnit.SyntaxTree;
         }
 
-        private static ImmutableArray<AttributeData>? GetAttributeData(CSharpCompilation compilation, SemanticModel document, CSharpSyntaxNode memberNode)
+        private static ImmutableArray<AttributeData>? GetAttributeData(Compilation compilation, SemanticModel document, SyntaxNode memberNode)
         {
             switch (memberNode)
             {
                 case CompilationUnitSyntax syntax:
-                    var validAttributesName = syntax.AttributeLists
-                                                    .SelectMany(x => x.Attributes)
-                                                    .Select(x => (x.Name as IdentifierNameSyntax)?.Identifier.ValueText)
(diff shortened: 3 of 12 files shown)


src/CodeGeneration.Roslyn/DocumentTransform.cs
```diff
@@ -51,16 +50,19 @@ public static class DocumentTransform
 
             var inputFileLevelUsingDirectives = inputSyntaxTree.GetRoot().ChildNodes().OfType<UsingDirectiveSyntax>();
 
-            var memberNodes = from syntax in inputSyntaxTree.GetRoot().DescendantNodes(n => n is CompilationUnitSyntax || n is NamespaceDeclarationSyntax || n is TypeDeclarationSyntax).OfType<MemberDeclarationSyntax>()
-                              select syntax;
+            var memberNodes = inputSyntaxTree.GetRoot().DescendantNodesAndSelf().OfType<CSharpSyntaxNode>();
```
amis92: Isn't removing descdentant filter going to cause traversal of every single syntax node below Type level? It's probably going to hurt performance badly, taking into account laziness of Roslyn's tree model; and I can't see why it'd ever be useful/necessary.

Pzixel: @amis92 this change was nessessary e.g. becaue `OfType<MemberDeclarationSyntax>` wasn't returning `CompilationUnitSyntax` or `NamespaceDeclarationSyntax` because they don't derive one from another. Another answer - user may want to attribute something else (e.g. property or field) and get generation based on it. Why this scenario should be not supported?

Code generation doesn't have often in my practice, there is often a separated assembly for all generated stuff, so it shouldn't regenerate a lot, however, it provides more flexibility. 

So my question: library should provide the widest support unless pefromance degradation isn't notable

amis92: Of course, I didn't contest `OfType` change, that I understand. Rather the `Descendants` change: right now we visit absolutely everything, including method bodies.

I think Properties and Fields might be good targets for Attributes, but on the other hand I believe a class that generates code should be decorated itself. And if it's a class that triggers generation, maybe better to leave the rest to the generator?

This is a big change from current behavior, and I think it's worth a separate discussion+PR.

Pzixel: Ok, it may be pushed as separated PR if/when we have need in it.

Pzixel:
What's about mergin this guy? I don't really like to create my own nuget source just to use this code. I have to write a generator that uses this feature today..

amis92:
Well, we'll have to wait for @AArnott here :) I think you'll have to setup custom nuget if you need it today, especially since releases here are scarce lately ;)

Pzixel:
@amis92 okay. Can you elaborate then, please?

I have built all packages into my private nuget (and have set version to 1.5 to pretend it's release), however I can't get it compiled - `dotnet exited with code 1`. Can you elaborate, please?

Here is my repo: https://github.com/Pzixel/Solidity.Roslyn .

I have built it with current branch and following `version.json`:
```json
{
  "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
  "version": "1.5",
  "publicReleaseRefSpec": [
    "^.+$", // we release out of master
  ],
  "cloudBuild": {
    "buildNumber": {
      "enabled": true
    }
  }
}
```

I have no idea what could be wrong...

amis92:
Please take a look here: https://github.com/AArnott/CodeGeneration.Roslyn/issues/54#issuecomment-350784608

The error is quite cryptic, and my best guess is because you do not reference DotNetCliTool in csproj. Of course the version would be what you set, 1.5. That's all I can think of.

Pzixel:
@amis92 you can see the csproj and see that I atually do reference it:
https://github.com/Pzixel/Solidity.Roslyn/blob/master/Solidity.Roslyn.Test/Solidity.Roslyn.Test.csproj#L1-L19

amis92:
I will be home soon, but from my experience it doesn't work well in xunit tests (at all). You'd have to have generated code in separate assembly referenced from test.

Pzixel:
@amis92 you didn't get it. It doesn't compile even if I exclude test project at all :) Maybe I'm just doing it wrong.

Pzixel:
@amis92 you don't need my nuget, you can build your own, it's just as simple as build this branch.

Pzixel:
Hi there! Any progress? I can provide prevompiled packages, if you have no possibility to pack them:
https://1drv.ms/u/s!Aho9_FLDH6NYg_wGpY3HjHyTQBK55g

Pzixel:
Hey everybody? Why it's so silent here?

@amis92 btw, you requested changes that I have already done. What's about marking it as fixed?

src/CodeGeneration.Roslyn.Tasks/GenerateCodeFromAttributes.cs
```diff
@@ -80,6 +83,9 @@ protected override string GenerateResponseFileCommands()
             argBuilder.AppendLine("--out");
             argBuilder.AppendLine(this.IntermediateOutputDirectory);
 
+            argBuilder.AppendLine("--projectDir");
+            argBuilder.AppendLine(ProjectDirectory);
```
AArnott: Please add the missing `this.` to match the style of the surrounding code.

src/CodeGeneration.Roslyn/CompilationGenerator.cs
```diff
@@ -104,6 +106,7 @@ public void Generate(IProgress<Diagnostic> progress = null, CancellationToken ca
                                 var generatedSyntaxTree = DocumentTransform.TransformAsync(
                                     compilation,
                                     inputSyntaxTree,
+                                    ProjectDirectory,
```
AArnott: leading `this.` prefix

src/CodeGeneration.Roslyn/TransformationContext.cs
```diff
@@ -13,20 +11,28 @@ public class TransformationContext
...
         {
             ProcessingMember = processingMember;
             SemanticModel = semanticModel;
             Compilation = compilation;
+            ProjectDirectory = projectDirectory;
         }
 
         /// <summary>Gets the syntax node the generator attribute is found on.</summary>
-        public MemberDeclarationSyntax ProcessingMember { get; }
+        public CSharpSyntaxNode ProcessingMember { get; }
```
AArnott: I don't think all nodes are "members", so this property could use a new name. And to avoid backward breaking changes, can you create a new member rather than change the existing one?

src/CodeGeneration.Roslyn/DocumentTransform.cs
```diff
@@ -112,22 +111,29 @@ public static class DocumentTransform
             return compilationUnit.SyntaxTree;
         }
 
-        private static IEnumerable<ICodeGenerator> FindCodeGenerators(SemanticModel document, SyntaxNode nodeWithAttributesApplied, Func<AssemblyName, Assembly> assemblyLoader)
+        private static ImmutableArray<AttributeData> GetAttributeData(Compilation compilation, SemanticModel document, SyntaxNode memberNode)
```
AArnott: rename `memberNode` to `syntaxNode`

src/CodeGeneration.Roslyn/TransformationContext.cs
Pzixel: All `MemberDeclarationSyntax` are `CSharpSyntaxNode`. Because in all practical examples they cast it to exact class it's backward-compatible, I guess. And I don't believe we should have to same nodes (referencing the same object) but where one property is of more specific type.

AArnott: Thanks. That's better.
Regarding the breaking change I mentioned, changing the type of a public member is a binary breaking change. But I guess for this project in particular, that's less interesting than a source breaking change. And we've made a few of those in the past, so I guess we're good.

Pzixel:
@AArnott Could you publish a new nuget version please? I thought it's configured to be automatically build by appveyor but it isn't.

AArnott:
It is published automatically, to an AppVeyor feed. Publishing to nuget.org is a separate step.
Yes, I can get that done today.

AArnott:
v0.4.37 is now on nuget.org

Pzixel:
@AArnott hello again.

For some reason, I'm not able to get this version. It fails with

> NU1102: Unable to find package CodeGeneration.Roslyn.Attributes with version (>= 1.0.0)
>   - Found 9 version(s) in api.nuget.org [ Nearest version: 0.4.37 ]
>   - Found 0 version(s) in Microsoft Visual Studio Offline Packages
>   - Found 0 version(s) in corefxlab

I'm not sure why it asks for 1.0.0, I didn't change anything related to versions in this PR. Can you elaborate, please?

amis92:
@Pzixel I've raised #68 for this.

PR #70, opened 2018-05-08
Inline System.CommandLine into Tool project, closes #63

### LINKED ISSUE #63: State of System.CommandLine
Are there any plans to change something on how the arguments are parsed.

If I see it correctly, the only reason the consumer of a generator written with this framework needs to put a `nuget.config` file in his project is because of how the arguments are parsed.

Looking at the repository, it does not seems like it will be out form beta (and on nuget) anytime soon.

I would like to have as less friction for the consumers of my libs as possible. Best case would be that only a nuget install is needed.

In my opinion the `nuget.config` could easily be avoided. 
+ Forking `System.CommandLine` and using the fork (until the original is out on nuget)
+ Using a different command line parser (don't know what's out there)
+ ~writing an own parser~

I think the first one would be a valid approach. Any thoughts on why this is a dumb idea?

amis92:
As stated in #63, the CoreFxLab System.CommandLine is no longer developed and will stay on MyGet forever, so I propose to just copy over the code from the CoreFxLab directly into CodeGeneration.Roslyn.Tool project.


AArnott:
Thanks for this!

PR #75, opened 2018-05-15
Logger

LokiMidgard:
When I was debugging (while building in VS) this framework, I had sometimes an bug which threw an exception. This was cached and an return code of 3 was the result.

Also different Console.Writeline were not shown. I Could have probably raised the output level of the build to verbose or even diagnostic, but this generates to much noise.

I have added a Logging class that allows to Log text at different loglevels to the build output.
In addition when returning a result of 3 now the exception will be logged to the output with highest log level.

### DIFF
--- a/src/CodeGeneration.Roslyn.Tasks/GenerateCodeFromAttributes.cs
+++ b/src/CodeGeneration.Roslyn.Tasks/GenerateCodeFromAttributes.cs
@@ -103,16 +103,26 @@ namespace CodeGeneration.Roslyn.Tasks
         {
             MessageImportance newImportance;
             if (DidExtractPrefix("High"))
+            {
                 newImportance = MessageImportance.High;
+            }
             else if (DidExtractPrefix("Normal"))
+            {
                 newImportance = MessageImportance.Normal;
+            }
             else if (DidExtractPrefix("Low"))
+            {
                 newImportance = MessageImportance.Low;
+            }
             else
+            {
                 newImportance = messageImportance;
+            }
 
             if (newImportance < messageImportance)
+            {
                 messageImportance = newImportance; // Lower value => higher importance
+            }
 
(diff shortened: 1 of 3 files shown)


src/CodeGeneration.Roslyn.Tool/Program.cs
```diff
@@ -53,8 +53,10 @@ static int Main(string[] args)
             {
                 generator.Generate(progress);
             }
-            catch
+            catch (Exception e)
             {
+                Logger.Log(LogLevel.High, $"{e.Message}: {e.Message}");
```
AArnott: You're emitting the exception message twice. Did you mean to log the exception type for the first one?

src/CodeGeneration.Roslyn.Tasks/GenerateCodeFromAttributes.cs
```diff
@@ -91,12 +91,41 @@ protected override string GenerateResponseFileCommands()
...
             }
 
             return argBuilder.ToString();
         }
+
+        protected override void LogEventsFromTextOutput(string singleLine, MessageImportance messageImportance)
+        {
+            MessageImportance newImportance;
+            if (DidExtractPrefix("High"))
+                newImportance = MessageImportance.High;
```
AArnott: Style: Can you please add `{ }` around all your single-line blocks?

src/CodeGeneration.Roslyn/Logger.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// Normal importance
+        /// </summary>
+        Normal = 1,
+
+        /// <summary>
+        /// Low importance, appears in more verbose logs
+        /// </summary>
+        Low = 2
+
```
AArnott: Can you add a trailing `,` after the last enum item?
Also move your blank line that is before the `}` to after it?

PR #79, opened 2018-06-21
DocumentTransform unit tests

### LINKED ISSUE #29: problem with #endregion
If I add Methods to an class using this generator, and at the end of the class is an ```#endregion```, this ```#endregion``` is also inserted in the generated code. This leads to an compiling error because ```#region``` is missing.

```c#
partial class Test {
    private bool test;
#region testRegion
[Generate]
void SomeMethod(){}
#endregion
}
```

Assuming the Generator does nothing this results in following:

```c#
partial class Test {
   
#endregion
}
```

### LINKED ISSUE #32: Problem with #if #endif in using
Hello,

If you have an #endif in end of using section Then you don't set it in the generated code

#if NETSTANDARD1_3
using SomeLib;
#else
using AnotherLib;
#endif

in generated code you get
#if NETSTANDARD1_3
using SomeLib;
#else
using AnotherLib;

so you are stuck

### Known workarounds
set an using after #endif

amis92:
Validates bugs #29, #32 as "expected"

These are unit tests for `DocumentTransform.TransformAsync` which, for a given "document" string validate if it generates document equal to "expected" document string.

In the current state they all pass. I think that it might be a better approach to first create a "working as-is" environment, and then add bugfixes separately.

---
*edit* after additional commits, this PR now:
* fixes #29
* fixes #32 
* adds unit tests that validate fixed behavior


amis92:
I'll happily fix these bugs, although I'm wondering what behavior *exactly* would we expect; which of the trivia should we keep, and which should we drop?

IMHO, we should drop all except `pragma`, `define` `if/else/endif`; those should probably stay, at least for `usings`.

Second thing is, should namespace-nested usings also be dropped (as is), or should we leave them too?

AArnott:
That all sounds good, with one response:

> Second thing is, should namespace-nested usings also be dropped (as is), or should we leave them too?

I don't care too much either way. I would tend toward including them. But as long as the code compiles I don't care too much. :)

amis92:
After reconsidering the implications, I'm going to drop all trivia, including if/else directives, regions and pragmas. Because generation is expected to be run for each compilation, and a given compilation has a defined set of `define`s; and *also* because if-deactivated code parts are trivia nodes in Roslyn Syntax model.

AArnott:
Does this actually fix #29 and #32 now (as opposed to writing tests to expect the prior behavior)?

amis92:
Yes, yes it does fix those two :) Glad to help.

Issue #40, opened 2017-07-06
Adding more control to output generation


Pzixel:
Currently I encountered two problems that are caused by current design:

1. Generated members always are nested in parent's namespace. Sometimes it's good, sometimes it's not required.
2. Generated members always are placed in same assembly where attribute is declared. Again, sometimes it's good, sometimes it's not required.

I propose to add more customisation here. I think ICodeGenerator have to provide somehow information if it want inherit ancestor's usings/namespace or it doesn't and information about path where it want to be stored:
```
Task<GenerationResult> GenerateAsync(TransformationContext context, IProgress<Diagnostic> progress, CancellationToken cancellationToken);

public class GenerationResult
{
   SyntaxList<MemberDeclarationSyntax> GeneratedMembers { get; }
   string DestinationProjectDirectoryFullPath { get; }
   bool IsNestedInAncestor { get; }
}
```

It's just an idea, I don't really like how it looks now, but we may invent something better. 

Use case: I have an attribute in a class in an assembly `foo` with multiple dependency, for example, ASP.Net. But I obviosly won't use it in my generator, so I don't want these dependencies. I can create assembly `foo.generated` without any dependency and store generation results there.

I can try to implement it if you agree that we need this feature and we can merge it in master. It's a bit more complicated than other changes and require some API rework, so I don't want to spend several days to get `I cannot afford merging it, it completly ruins my design` :)

AArnott:
Thanks for checking first. I need to mull this over a bit before advising you to proceed. 
The most concern comes from creating an entirely new assembly for the generated. That is rather non-trivial because it turns out there are lots of concerns that spring up when you do that, like satellite assemblies, signing, added references, etc. I don't think I want to go there without at least one more use case from someone else to justify it. But I do appreciate you offering to do the work.

As for the ability to generate members in another namespace, that sounds like a bad idea in general, but for specific use cases may be perfectly legit (for example, if you're writing a generator specialized for your own purposes rather than for general reuse). I would accept a PR that makes this possible. And I like how your proposed design will allow for enhancements in the future in the generation result without breaking changes to existing generators (although this one itself will be a breaking change).

Pzixel:
@AArnott completly different assembly may led to other multiple problems (such as how to reference it?). I think empty assembly for generated classes is quite OK. You can sign it, add versioning, and so on, using common flow. Implementing it as result of generation may be complicated and unwanted.

AArnott:
[...]

Issue #30, opened 2017-06-12
Allow to add custom usings in generated code.


bhugot:
Hello, I met a problem as i'm trying to generate code for NET3.5 and netstandard for Diagnostics.Tracing but in NET3.5 I have Microsoft.Diagnostics.Tracing and for netstandard it's System.Diagnostics.Tracing.

I don't know if there is a way so that the generator get the targetFramework or if I could add some using to my generated class.

Pzixel:
@bhugot I'm currently using this workaroung:

```
var clientsNamespace = NamespaceDeclaration(ParseName("Clients"))
	.AddUsings(
		GetUsingDirectiveSyntax("System"),
		GetUsingDirectiveSyntax("System.Collections.Generic"),
		GetUsingDirectiveSyntax("System.Threading.Tasks"))
	.AddMembers(clientClass);

return Task.FromResult(List<MemberDeclarationSyntax>().Add(clientsNamespace));
```
It adds additional namespace but it's ok in my case. Probably, it helps you.

bhugot:
@Pzixel Thanks found another way to do it for my case.

Pzixel:
@bhugot then, please, share it? :) Maybe it would fit me better.

bhugot:
I just use fullName for each class.

amis92:
#80 resolved this.

PR #80, opened 2018-06-24
Rich generators

amis92:
**It's just an initial concept draft to start up a discussion**

I've got an idea for a big enhancement: a feature that would enable many more scenarios for code generation in this framework, namely:

* add ability to completely control ancestry of generated code (namespace, ancestor partials) (#40)
* add top-level (compilation unit) usings (#30)
* add top-level extern aliases
* add top-level attributes (`[assembly: Xyz]`)

My proposal adds a new interface that has a method with the same parameter list as current `ICodeGenerator.GenerateAsync`, but returns `Task<RichGenerationResult>` which contains four syntax lists: Members, Usings, Externs and AttributeLists that get merged with those created by previous generators and then are added to the generated compilation unit.

There are no tests currently, but I will add them, if the design would be accepted.


amis92:
@AArnott I've rebased onto master after merging #79. What do you think of the proposed enhancements?

amis92:
@AArnott ping? What are your thoughts?

amis92:
@AArnott pinging again. Any reaction?

src/CodeGeneration.Roslyn/CodeGeneration.Roslyn.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard1.6</TargetFramework>
+    <LangVersion>7.1</LangVersion>
```
AArnott: Let's go ahead and shoot for 7.3. :smile:

amis92:
@AArnott I've added tests, done some refactors and bumped lang ver to 7.3 as requested.

I think I'm done here. What's your take on this?

amis92:
One major refactor I should maybe explain is the move of wrapping-member-in-ancestor logic; I've moved it into a separate nested class which I've named `EnrichingCodeGeneratorProxy`. It essentially hides that logic behind `IRichCodeGenerator` interface.

So the processing is now more streamlined: non-rich generators are now wrapped into Proxy class, and only the Rich generation is called in the `TransformAsync` foreach loop.

One other thing I've been uncertain, is: should the `TransformationContext` and `RichGenerationResult` have public constructors? Maybe we could offer some factory methods instead. That way we could reserve space for future changes/additions without breaking existing code.

amis92:
@AArnott pinging for review :)

AArnott:
Sorry... looking now..

src/CodeGeneration.Roslyn/DocumentTransform.cs
```diff
@@ -240,5 +206,75 @@ private static string GetFullTypeName(INamedTypeSymbol symbol)
 
             return nameBuilder.ToString();
         }
+
+        private class EnrichingCodeGeneratorProxy : IRichCodeGenerator
+        {
+            public EnrichingCodeGeneratorProxy(ICodeGenerator codeGenerator)
```
AArnott: >codeGenerator [](start = 62, length = 13)

Since we dereference this later without checking, please add `Requires.NotNull(codeGenerator, nameof(codeGenerator));` to this ctor.

AArnott:
I don't see how a factory method will make it more future-proof than a ctor. We would presumably add more of either one as we add members. However, the fact that you're already using default parameters on the first "overload" may make it tricky to add more overloads with default parameters later. I'll leave more comments in the relevant files.

---
In reply to: [413479597](https://github.com/AArnott/CodeGeneration.Roslyn/pull/80#issuecomment-413479597) [](ancestors = 413479597)

src/CodeGeneration.Roslyn/RichGenerationResult.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="usings">Assigned to <see cref="Usings"/>.</param>
+        /// <param name="attributeLists">Assigned to <see cref="AttributeLists"/>.</param>
+        /// <param name="externs">Assigned to <see cref="Externs"/>.</param>
+        [DebuggerStepThrough]
+        public RichGenerationResult(
+            SyntaxList<MemberDeclarationSyntax> members,
+            SyntaxList<UsingDirectiveSyntax> usings = default,
+            SyntaxList<AttributeListSyntax> attributeLists = default,
+            SyntaxList<ExternAliasDirectiveSyntax> externs = default)
+        {
```
AArnott: Instead of these ctors, which will be troublesome to version as we add more members, why not give the properties public setters and let them set whatever ones they want?

src/CodeGeneration.Roslyn/RichGenerationResult.cs
```diff
@@ -0,0 +1,67 @@
...
+            out SyntaxList<UsingDirectiveSyntax> usings,
+            out SyntaxList<AttributeListSyntax> attributeLists,
+            out SyntaxList<ExternAliasDirectiveSyntax> externs)
+        {
+            members = Members;
+            usings = Usings;
+            attributeLists = AttributeLists;
+            externs = Externs;
+        }
+    }
```
AArnott: I'm not a fan of deconstruction unless it's *very* obvious what the members will be to the caller and what order they should come in (i.e. first and last name). As the members grow, their order in this list are basically arbitrary -- they look arbitrary right now. That leads to bugs in code when you add/refactor in this class but fail to update every caller (which is hard, since they don't actually call Deconstruct explicitly) and it happens to compile because the types between two reordered members remains the same.
It also requires that the caller explicitly name the variables when it would produce smaller, more readable code if they just dereferenced the property as it is defined in this struct each time.

src/CodeGeneration.Roslyn/TransformationContext.cs
```diff
@@ -12,13 +17,19 @@ public class TransformationContext
...
         /// <param name="compilation">The overall compilation being generated for.</param>
         /// <param name="projectDirectory">The absolute path of the directory where the project file is located</param>
+        /// <param name="compilationUnitUsings">The using directives already queued to be generated.</param>
+        /// <param name="compilationUnitExterns">The extern aliases already queued to be generated.</param>
         public TransformationContext(CSharpSyntaxNode processingNode, SemanticModel semanticModel, CSharpCompilation compilation,
-                                     string projectDirectory)
+                                     string projectDirectory,
+                                     IEnumerable<UsingDirectiveSyntax> compilationUnitUsings,
+                                     IEnumerable<ExternAliasDirectiveSyntax> compilationUnitExterns)
         {
```
AArnott: While this is a technically breaking change, it should only break unit tests that may have been written for generators, which isn't as severe.
We could make this internal, but that would make mocking up those unit tests even harder. So I'm good with this change.

amis92:
I'm glad I could help! All the comments were valid and I see the point. I've got a little carried away on that immutability bandwagon. :)

PR #82, opened 2018-07-11
Update README for v0.4.49

amis92:
* Updated required target framework versions
* Removed references/requirements of custom NuGet feed for CoreFxLab MyGet
* Added VS2017 and SDK-csproj requirement notes

Also added Table of Contents to the top.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,20 +8,30 @@ This includes design-time support, such that code generation can respond to
 changes made in hand-authored code files by generating new code that shows
 up to Intellisense as soon as the file is saved to disk.
 
+## Table of Contents
+
+* [How to write your own code generator][]
+* [Apply code generation][]
+* [Developing your code generator][]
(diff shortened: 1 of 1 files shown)


amis92:
I checked, they do work. Either there's somewhere a problem with rendering, or you simply missed how that form of link references works :)

https://github.com/amis92/CodeGeneration.Roslyn/blob/92e1588989818629bda115658fc2b3216e36f65a/README.md

BTW, I can't get NuGet package badge to render, it's always a broken image for me. Do you have that problem too?

LokiMidgard:
For me NuGet badge is rendered (through it takes some time)

AArnott:
I see. I've never seen that syntax before and when I tried it in GitHub's rendered diff view, it took me to a 404. But I see in the normal context it works. 
Thanks.

PR #91, opened 2018-09-06
Split Build Task into Prepare and Core targets

amis92:
Added Condition on Core target to prevent "returned 1" error (no Compile files)


amis92:
@AArnott pinging for comment/review

AArnott:
Thanks.

PR #92, opened 2018-09-06
Update Roslyn version to 2.9.0, require exact in NuGet

### LINKED ISSUE #77: Please update your CSharp reference
The ``Microsoft.CodeAnalysis.CSharp`` reference is [out of date](/AArnott/CodeGeneration.Roslyn/blame/8113bbd8f705b56653ee94bb0e870cb05cfa5001/src/CodeGeneration.Roslyn/CodeGeneration.Roslyn.csproj#L20) and seems most likely the culprit as to why my [Generators](/mwpowellhtx/Kingdom.Collections/tree/master/src/Kingdom.Collections.Enumerations.Generators) are not working.

Otherwise, my generators work just fine for me [under unit test](/mwpowellhtx/Kingdom.Collections/tree/master/src/Kingdom.Collections.Enumerations.Generators.Tests); that is, I can [gen up a TransformationContext](/mwpowellhtx/Kingdom.Collections/blob/master/src/Kingdom.Collections.Enumerations.Generators.Tests/FlagsEnumerationGeneratorTests.cs#L191), [work with my own Roslyn Compilation result](/mwpowellhtx/Kingdom.Collections/blob/master/src/Kingdom.Collections.Enumerations.Generators.Tests/FlagsEnumerationGeneratorTests.cs#L189), and [Generate a member list](/mwpowellhtx/Kingdom.Collections/blob/master/src/Kingdom.Collections.Enumerations.Generators.Tests/FlagsEnumerationGeneratorTests.cs#L113). No problem.

Sidebar, ideally, breaking changes notwithstanding, *NuGet* should respect the ``>=base-version``, but if there is a specific version, [you can specify a version range](https://docs.microsoft.com/en-us/nuget/reference/package-versioning#version-ranges-and-wildcards) in your [``<PackageReference/>``](/AArnott/CodeGeneration.Roslyn/blob/master/src/CodeGeneration.Roslyn/CodeGeneration.Roslyn.csproj#L20). The same range syntax applies as with NuSpec file format, apparently. To what versions, however, I cannot say, and is a matter you need to work out.

Minimally I would suggest a two pronged update for you. First, identify the breaking change, use that version as a max version in the range, and publish that update. That would at least make it obvious for folks like myself wanting to do code generation that there is a hard version constraint. Then, see if you cannot track to the latest version, and provided you can, or without too much difficulty (i.e. those breaking change(s), mind you), publish that update following up.

In the meantime, I need to see if I can reference a version of that package; hopefully it doesn't break too much in my own code that I am depending upon or have otherwise made assumptions around.

amis92:
`Directory.Build.props` gained `RoslynNugetVersion` with ~~`2.*`~~ `2.9.0` value ~~to always depend on latest Roslyn.~~

`CodeGeneration.Roslyn` has strict dependency on this version: `Version="[$(RoslynNugetVersion)]"`

Closes #77

### DIFF
--- a/src/CodeGeneration.Roslyn.Tests/CodeGeneration.Roslyn.Tests.csproj
+++ b/src/CodeGeneration.Roslyn.Tests/CodeGeneration.Roslyn.Tests.csproj
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.2.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="$(RoslynNugetVersion)" />
     <PackageReference Include="Microsoft.CSharp" Version="4.4.1" />
     <PackageReference Include="Xunit" Version="2.3.1" />
     <PackageReference Include="Xunit.runner.visualstudio" Version="2.3.1" />
--- a/src/CodeGeneration.Roslyn/CodeGeneration.Roslyn.csproj
+++ b/src/CodeGeneration.Roslyn/CodeGeneration.Roslyn.csproj
@@ -18,7 +18,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.0.4" />
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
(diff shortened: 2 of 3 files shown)


amis92:
I've been misled by MSBuild/deps.json cache to believe it could work with floating version - it cannot. I've since fixed my error, and made the `RoslynNugetVersion` be exact, currently `2.9.0`.

amis92:
@AArnott pinging for comment/review

AArnott:
Thanks.

PR #93, opened 2018-09-06
Remove old dependency resolution files

amis92:
I think these are from some old incarnation, surely currently unused.


AArnott:
Oops. I should have rejected this. See #103

PR #94, opened 2018-09-06
Feature: split main package into Engine and core

amis92:
Engine package contains the `DocumentTransform` and `CompilationGenerator` classes, taking the responsibility of all the operations.

Main package `CodeGeneration.Roslyn` keeps the public API surface required for building generators - Generator interfaces, Context and Result types.

I've also added solution folders, because single level list started to be crowded, but that's purely optional and obviously a little off-topic. There is "tests" folder with all test projects, "lib" folder for all libraries, and the tool app is kept at the root level.


amis92:
This PR will be rebased onto actual master when/if #91, #92, #93 get merged/dropped.

amis92:
@AArnott I've dropped the DeclarationComputer files *again*, but this time I've created https://github.com/AArnott/ImmutableObjectGraph/pull/113 so that ImmutableObjectGraph keeps working with future releases of this project.

I've rebased the changes on to the current master, so that the diff shows correctly.

The aim of this PR is still the same as it's declared in OP.

What do you think?

amis92:
Oh, also I've bumped the version to 0.5 to indicate a major change. I think that's ok?

amis92:
@AArnott ping? 🎵 *Another month bites the dust* 🎵

src/CodeGeneration.Roslyn.Tool/Program.cs
```diff
@@ -1,11 +1,15 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
-using CodeGeneration.Roslyn.Tool.CommandLine;
-using Microsoft.CodeAnalysis;
+﻿// Copyright (c) Microsoft. All rights reserved.
```
AArnott: This isn't a Microsoft copyrighted work.

AArnott:
I love your little tune. :) I laughed a long time at that.

Sorry for the delay.

src/CodeGeneration.Roslyn.Engine/CodeGeneration.Roslyn.Engine.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard1.6</TargetFramework>
+    <LangVersion>7.3</LangVersion>
+    <SignAssembly>True</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\opensource.snk</AssemblyOriginatorKeyFile>
+    <Description>The engine of source code generation used by `dotnet-codegen` tool.</Description>
```
AArnott: So I guess this creates a new nuget package. Is that right? And does the distribution all work out, including with the `dotnet-codegen` tool?

AArnott:
Oh, I guess the test failures are a pre-existing issue. So if you can just resolve the two comments I made to your changes we can complete this. :)

src/CodeGeneration.Roslyn.Engine/CodeGeneration.Roslyn.Engine.csproj
amis92: Yes, I can build a generator-using project from local NuGet feed with these artifacts.

In terms of pushing this new package to official NuGet or MyGet feeds, I don't know how that'll work.

src/CodeGeneration.Roslyn.Tool/Program.cs
amis92: Fixed

PR #103, opened 2018-10-07
Revert "Remove old dependency resolution files"

AArnott:
Reverts AArnott/CodeGeneration.Roslyn#93

I'm not sure why #93 called them "old dependency resolution files", but these classes have nothing to do with dependency resolution, and in fact they were public API that my ImmutableObjectGraph generator requires. These classes came from roslyn, where they are internal.


AArnott:
I added them a few years ago. I'm not sure why they're here as opposed to in the generator extension assemblies -- there may be a good reason, but I don't recall.

amis92:
@AArnott I think it'd be best to drop them from CodeGeneration.Roslyn (and move to ImmutableObjectGraph?). They don't serve a purpose here, aren't tested, add unnecessary public API surface and are completely separate from the actual purpose and features of this project. At least that's what I believe to be true.

It may be a longer process of deprecation/moving over, but since the project is v0.4, breaking changes like that are better than when we have 1.x :)

PR #107, opened 2018-10-23
Resolve NuGet assemblies from path file name in addition to package name

### LINKED ISSUE #106: How to use additional dependencies
In order to create some syntax nodes I'm trying to add more NuGet dependencies to my custom generator project. More specifically I want to use some methods provided by [CSharpSyntaxGenerator.cs](https://github.com/dotnet/roslyn/blob/master/src/Workspaces/CSharp/Portable/CodeGeneration/CSharpSyntaxGenerator.cs) from the [Microsoft.CodeAnalysis.CSharp.Workspaces](https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.Workspaces/) NuGet package - such as [TypedConstantExpression(TypedConstant value)](https://github.com/dotnet/roslyn/blob/5a8dabf92d3abb17b47b9c54d33d93313dc8d725/src/Workspaces/CSharp/Portable/CodeGeneration/CSharpSyntaxGenerator.cs#L3937).

When I test the code generator from unit tests it works perfectly well.
When I'm running the build on a project which uses the generator it fails with the following error:
```
2>Exception in file processing: System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.CodeAnalysis.Workspaces, Version=2.9.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
2>File name: 'Microsoft.CodeAnalysis.Workspaces, Version=2.9.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
```
My understanding is that dotnet-codegen will only load the assemblies using this [resolver](https://github.com/AArnott/CodeGeneration.Roslyn/blob/3109bcd51df4f16881881562b09f5785105c12da/src/CodeGeneration.Roslyn/CompilationGenerator.cs#L83). I don't know how to make sure that the additional dependencies I'm adding to my generator will work when executed by  `<DotNetCliToolReference Include="dotnet-codegen" Version="0.4.88" />`.

My questions:
1. When I'm using my custom generator as a .csproj project (ie. not as a NuGet dependency), how do I make sure that the additional NuGet references added to this project will be resolved by `<DotNetCliToolReference Include="dotnet-codegen" Version="0.4.88" />`?
2. Will this problem be resolved anyways if I use my generator as a NuGet package instead?
(as the additional references would then be managed properly)
3. All I need is using [Microsoft.CodeAnalysis.CSharp.Workspaces](https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.Workspaces/) NuGet package. Would it make sense to add this dependency in CodeGeneration.Roslyn as there are already many dependencies with Roslyn? I can create a PR.

frblondin:
Attempt to fix #106 

Not sure whether a unit test is required/easy to add...

### DIFF
--- a/src/CodeGeneration.Roslyn/CompilationGenerator.cs
+++ b/src/CodeGeneration.Roslyn/CompilationGenerator.cs
@@ -209,7 +209,7 @@ namespace CodeGeneration.Roslyn
 
         private Assembly ResolveAssembly(AssemblyLoadContext context, AssemblyName name)
         {
-            var library = this.dependencyContext.RuntimeLibraries.FirstOrDefault(runtime => string.Equals(runtime.Name, name.Name, StringComparison.OrdinalIgnoreCase));
+            var library = FindMatchingLibrary(this.dependencyContext.RuntimeLibraries, name);
             if (library == null)
                 return null;
             var wrapper = new CompilationLibrary(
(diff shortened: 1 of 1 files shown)


frblondin:
Is there any chance this PR can be reviewed?

Thanks

PR #117, opened 2019-01-24
Overall refactor, and license streamlining

### LINKED ISSUE #116: Warning NU5125: The 'licenseUrl' element will be deprecated. Consider using the 'license' element instead.

amis92:
Applies StyleCop as required, adjusted to the project, and applied some suggestions.

Some more Directory.Build generalization, refactor, properties split into labeled groups.

Fixes #116

No functional changes.


src/Directory.Build.props
```diff
@@ -1,29 +1,47 @@
...
+
+  <PropertyGroup Label="Build configuration">
+    <LangVersion>7.3</LangVersion>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)..\obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
     <OutputPath>$(MSBuildThisFileDirectory)..\bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
     <PackageOutputPath>$(MSBuildThisFileDirectory)..\bin\Packages\$(Configuration)\</PackageOutputPath>
+    <IsTestProject>$(MSBuildProjectName.Contains('Test'))</IsTestProject>
+    <IsPackable Condition=" $(IsTestProject) ">false</IsPackable>
+    <GenerateDocumentationFile Condition=" !$(IsTestProject) ">true</GenerateDocumentationFile>
```
AArnott: We should *always* generate a documentation file. Otherwise C# doesn't run its analysis on the xml doc comments that *do* exist and flag syntax and reference errors. 
It's fine though to use `<NoWarn>$(NoWarn);CS1591</NoWarn>` in test projects so that no documentation is *required* in test projects though. 
I believe there is even a StyleCop rule that calls out this pattern. Is it disabled?

src/Directory.Build.props
```diff
@@ -1,29 +1,47 @@
...
+    <CodeAnalysisRuleSet Condition=" !$(IsTestProject) ">$(MSBuildThisFileDirectory)/CodeGeneration.Roslyn.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRuleSet Condition=" $(IsTestProject) ">$(MSBuildThisFileDirectory)/CodeGeneration.Roslyn.Tests.ruleset</CodeAnalysisRuleSet>
+    <WarningsAsErrors>true</WarningsAsErrors>
```
AArnott: Not here, please. Warnings should *not* break the build when we're in the "inner loop" of development. We want this set for cloud builds (which we can do via environment variable or msbuild parameter in the .yml file), and if you want we could set this here, conditional on release builds.

src/Directory.Build.props
```diff
@@ -1,29 +1,47 @@
...
-    <IsPackable Condition=" $(MSBuildProjectName.Contains('Test')) ">false</IsPackable>
-    <NerdbankGitVersioningVersion>2.2.13</NerdbankGitVersioningVersion>
+    <LicenseExpression>MS-PL</LicenseExpression>
+  </PropertyGroup>
+
+  <PropertyGroup Label="Shared versions">
     <RoslynNugetVersion>2.9.0</RoslynNugetVersion>
   </PropertyGroup>
 
+  <PropertyGroup Label="Signing" Condition=" !$(IsTestProject) ">
```
AArnott: I'm not a fan of all the labeling. It becomes too hard to predict. For example, why wouldn't signing be part of your "build configuration" group? I'd only know to put or look for them elsewhere after reviewing every labeled group. "NuGet metadata" *seems* clear, but then `Copyright` and `Company` *also* impact the assembly attributes that are generated and compiled into the assembly, so they aren't even themselves just nuget metadata.

amis92:
@AArnott can you please re-review? I've applied changes per your comments, but still am not sure about the Directory.Build.props organization.

src/Directory.Build.props
amis92: I've applied the suggestion.

src/Directory.Build.props
amis92: I added condition for "Configuration" == "Release". Not sure if that's enough in your opinion.

src/Directory.Build.props
amis92: I removed all except one for the "Metadata". Not sure about the name, I've thought about "Author(ing)", or maybe I should just merge it into the main group.

amis92:
I've merged PropertyGroups but split out three:
* Author/Metadata
* Non-test-project properties
* Test-project properties

How do you feel about that?

PR #126, opened 2019-04-10
Target netstandard2.0 and use Roslyn 3.0

### LINKED ISSUE #124: Catching up the dotnet standard
Hello,

What is the constraint that is hindering the ``netstandard1.6`` versus ``netstandard2.0`` target frameworks?

I ask because the latest (experimental? or at least a couple of the dependency dependencies are in beta, it seems) references to [``Microsoft.CodeAnalysis.CSharp``](http://nuget.org/packages/Microsoft.CodeAnalysis.CSharp) are incompatible with ``netstandard1.6``.

For now, I'll just walk back the reference dependency versions, but it might be a migration path issue moving forward?

Thanks!

amis92:
So, since the project right now is completely independent of Visual Studio version, I've upgraded Roslyn to v3.0 and targeted netstandard2.0, and the tool itself now targets netcoreapp2.1.

The only requirements are:
* SDK-style csproj for DotNetCliTool to resolve and
* .NET Core SDK 2.1+ to run it

Thanks to CodeGeneration.Roslyn not integrating with MSBuild now, the tool runs as a .NET Core app completely independent of all the other parts (VS, MSBuild, platform etc). The price is that we always use our own Roslyn version which we need to update manually.

Triggered by, and closes #124

### DIFF
--- a/src/CodeGeneration.Roslyn.Tests.Generators.Dependency/CodeGeneration.Roslyn.Tests.Generators.Dependency.csproj
+++ b/src/CodeGeneration.Roslyn.Tests.Generators.Dependency/CodeGeneration.Roslyn.Tests.Generators.Dependency.csproj
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard1.6</TargetFramework>
-    <PackageTargetFallback>$(PackageTargetFallback);portable-net45+win8+wp8+wpa81;</PackageTargetFallback>
   </PropertyGroup>
 
 </Project>
--- a/README.md
+++ b/README.md
@@ -23,24 +23,27 @@ your code generation attribute is applied to, but with a suffix appended to its 
 
 ### Prerequisites
 
-* .NET Core SDK v2.x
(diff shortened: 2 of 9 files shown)


amis92:
Also, @AArnott - after it's merged, can you create a release? :)

src/CodeGeneration.Roslyn.Tests.Generators.Dependency/CodeGeneration.Roslyn.Tests.Generators.Dependency.csproj
```diff
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard1.6</TargetFramework>
```
Pzixel: Is it intended to leave 1.6 here?

amis92: Yes, since it's a dependency *of* the generator, it can be left, it shouldn't matter.

Issue #150, opened 2019-06-30
[RFC] Attribute delivered as nuget contentFiles


amis92:
Recently I've been thinking about how to un-tfm BuildTime and Attributes packages.

Since the move to a separate netcoreapp-based tooling, we no longer require any specific type of framework to be compatible with (did we at any point in time?). As such, the TFM in BuildTime is only unnecessarily complicating things. That package should not declare any frameworks, and just deliver the msbuild files.

Attributes package theoretically *does* contain a `dll` used and referenced almost anywhere. But in reality, all it has is a single Attribute file, it's unique identity is not used anyway (just a simple `Type.Name` comparison, doesn't even account for the namespace):

https://github.com/AArnott/CodeGeneration.Roslyn/blob/2b937a442a1380eb4c65fa7da45077ba62a10b33/src/CodeGeneration.Roslyn.Engine/DocumentTransform.cs#L154-L162

This single file could actually be "inlined" as a compiled source file in a generator-attribute containing project. It'd remove dependency on a whole DLL, remove any framework checks, etc. For published generators, the attributes package could also use the same philosophy to remove the unnecessary package dependencies.

What are your feelings, @AArnott ?

AArnott:
Not a fan. I've done it both ways a few times. Including source files into a project is problematic as different stylecop rules, compiler settings and other analyzers tend to make someone unsatisfied with the experience. Also as far as the assembly dependency, it never loads for attributes except when reflection queries for them. In fact we can easily prevent the attributes from surviving in the final assembly by attributing our attribute with a ConditionalAttribute like this: https://github.com/dotnet/corefx/blob/da3ba7ebe34771dde769b65dfb7f78fa55a5abb6/src/Common/src/CoreLib/System/Diagnostics/CodeAnalysis/SuppressMessageAttribute.cs#L23

That, and compiling our attributes assembly to netstandard1.0, should avoid all trouble.

amis92:
Agreed. Did not take into account, well, most of what you said. :)

PR #151, opened 2019-07-01
TFM-agnostic BuildTime package, Attributes with less TFMs

amis92:
Based on #150, I've decided to:
* greatly simplify TFM support for Attributes to just `netstandard` `1.0` ~~and `2.0`~~.
* make BuildTime package TFM-agnostic since it contains no DLLs

Net result: much quicker builds, smaller package size, almost enables x-plat building


src/CodeGeneration.Roslyn.Attributes/CodeGeneration.Roslyn.Attributes.csproj
```diff
@@ -1,14 +1,7 @@
...
   <PropertyGroup>
-    <TargetFrameworks>
-      netstandard1.0;
-      netstandard2.0;
-      net40;
-      net20;
-      portable-net40+win8+wpa81;
-      portable-net45+win8+wpa81+wp8
-    </TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;netstandard2.0</TargetFrameworks>
```
AArnott: Why is 2.0 needed? This would double the compile time of this assembly and I can't think of why we'd need it.

src/CodeGeneration.Roslyn.BuildTime/CodeGeneration.Roslyn.BuildTime.csproj
```diff
@@ -1,19 +1,18 @@
-﻿<Project Sdk="MSBuild.Sdk.Extras">
+﻿<Project Sdk="Microsoft.Build.NoTargets">
```
AArnott: Cool! I didn't know about this. But if this project has no outputs, why do we need it? Are we packing with it even though it doesn't produce an assembly? If so, "NoTargets" sounds like a very misleading name.

amis92: Well, yeah - it clears the "Build" target AFAIK, but Pack still works, so I used that :) So, no assembly, no compilation (considerable time saved), but package is still produced.

The name is a bit misleading indeed, I didn't notice really because I've been mostly reading the actual docs, not names 😅

src/CodeGeneration.Roslyn.Attributes/CodeGeneration.Roslyn.Attributes.csproj
amis92: Yes, the compile time is doubled.

I've kept netstandard2.0, because it has a much smaller dependency footprint for dependents.

Compare
* NETStandard2.0 deps list for v2.0.3: https://www.nuget.org/packages/NETStandard.Library/2.0.3
* NETStandard1.6 deps list for v1.6.1: https://www.nuget.org/packages/NETStandard.Library/1.6.1

Standard2.0 is much more lightweight. However, I'm fine with dropping that target. I just did it because I thought it could provide a better experience for consumers. (I think referencing netstandard2.0 in addition to any 1.x is a kind of "magic" I learned by browsing other projects - it mey very well be a [Cargo Cult](https://en.wikipedia.org/wiki/Cargo_cult) by now!)

AArnott: If the attribute package doesn't actually express that netstandard.library dependency it shouldn't be an issue. And since with the right condition on our attribute even a dependency on itself evaporates, I don't see any advantages to adding a netstandard2.0 target.

amis92: Ok, I've gone with 1.0 only

PR #152, opened 2019-07-01
Add GeneratorInConsumerSolution sample

amis92:
Finally, an internal sample to show


samples/GeneratorInConsumerSolution/GeneratorInConsumerSolution.sln
```diff
@@ -0,0 +1,48 @@
...
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sample.Consumer", "Sample.Consumer\Sample.Consumer.csproj", "{E47BD7FC-FD4A-44BF-AA67-787CDE3C8A21}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sample.Generator", "Sample.Generator\Sample.Generator.csproj", "{F15F35D6-932C-40BB-9CB9-56EC6F1942AE}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
+		Release|Any CPU = Release|Any CPU
```
AArnott: nit: Can we remove the x86 and x64 solution platforms?

AArnott:
The `samples/GeneratorInConsumerSolution/README.md` is claimed to be binary by github. Did you encode it with UTF-16, perhaps? Please re-encode as UTF-8.

samples/GeneratorInConsumerSolution/nuget.config
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
+    <clear />
+    <add key="local CG.R Debug packages" value="../../bin/Packages/Debug" />
```
AArnott: Using Debug packages unconditionally is unfortunate. This would fail if someone (including the build CI if it starts validating this) built Release, or (worse) cause unexpected results if they built Debug a long time ago, made changes and cleared the nuget cache, and built in release and still saw the old debug behavior.
I believe nuget sources can be added as msbuild items, and while I generally hate that pattern, in this case it might make sense so that you can add the path with `$(Configuration)` used.

amis92: Good suggestion! Found it: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#restore-properties

amis92: I've tried but found a problem: https://github.com/AArnott/CodeGeneration.Roslyn/pull/152/commits/a948f29212c11f2f4c5c482e647368e1eefbf49b

DotNetCliToolReference isn't working with that scenario, at least I suspect that's the cause.

samples/GeneratorInConsumerSolution/Directory.Build.props
```diff
@@ -1,7 +1,9 @@
 <Project>
+
+  <Import Project="$(MSBuildThisFileDirectory)../PackageSources.props"/>
   
   <PropertyGroup>
-    <LocalNuGetVersion>*</LocalNuGetVersion>
+    <LocalNuGetVersion>0.6.5-g5df32cb0b6</LocalNuGetVersion>
```
AArnott: Was this change intentional?

samples/GeneratorInConsumerSolution/nuget.config
AArnott: OK, well thanks for trying. What you have here will work most of the time. :)

samples/GeneratorInConsumerSolution/Directory.Build.props
amis92: Definitely not 😨

amis92: I'll revert the whole commit anyway, then.

PR #174, opened 2019-11-21
.NET Core 3 support dotnet-codegen

### LINKED ISSUE #169: .NET Core 3/netstandard 2.1 support?
Should it work?
After upgrading my projects, which work fine with .NET Core 2.1 and codegen 0.6.1, any project with codegen fails to compile, complaining about locating "netstandard, Version 2.1.0.0, ...", with CodeGeneration.Roslyn at the top of the call stack.

kuhsaft:
Create runtimeconfig.template.json containing rollForward for .NET Core 3 support

This is what was done for FAKE: https://github.com/fsharp/FAKE/issues/2325
Fixes #169

### DIFF
--- a/src/CodeGeneration.Roslyn.Tool/runtimeconfig.template.json
+++ b/src/CodeGeneration.Roslyn.Tool/runtimeconfig.template.json
@@ -1,0 +1,3 @@
+{
+    "rollForward": "Major"
+}


amis92:
For future reference the actual switch documentation that I could find is at https://docs.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-2-1#roll-forward (introduction doc)

amis92:
Reading futher, this is the design document about runtime bindings:
https://github.com/dotnet/designs/blob/master/accepted/runtime-binding.md#rollforward

This article specifies that the rollForwardOnNoCandidateFx is deprecated starting with 3.0, and RollForward is the recommended one.

I'd like to go with the RollForward in that case - and it can be specified within `csproj` when using SDK 3.0.

kuhsaft:
I changed `rollForwardOnNoCandidateFx` to `rollForward` in `runtimeconfig.template.json` since that's what recommended in .NET Core 3. The csproj property doesn't work unless you build using .NET Core 3.

amis92:
I've used RollForward property with SDK v3.1 - see #178

PR #175, opened 2019-11-26
Support C# Code Snippets to generate code.

Corniel:
Using code snippets as starting point for code generation can make the code generator much easier to read/maintain. Writing code with the `SyntaxTreeFactory` can be hard.

So how does should this work?

We have some file (embedded, or disk)
``` C#
public struct @Name
{
    public @Name(int count) => Count = count;
    public int Count { get; }
}
```

With some simple generator files. What kind of transform you want to apply should be as free as possible, so let the generator decide for its own. In this example its a simple string replace.
``` C#
class SimpleCodeSnippet : CSharpCodeSnippet<TransformArguments>
{
    public SimpleCodeSnippet(string text)
        : base(text, "test.cs", null) { }

    protected override string TransformText(TransformArguments arguments)
    {
        return Text.Replace("@Name", arguments.Name);
    }
}
class TransformArguments
{
    public string Name { get; set; }
}
````

This can be used like this:
``` C#
SimpleCodeSnippet snippet = GetCodeSnippet(); // embedded resource, disk, whatever.
var syntax = await snippet.ParseAysnc<StructDeclarationSyntax>(new TransformArguments { Name = "TestStruct" });
```

### DIFF
--- a/src/CodeGeneration.Roslyn.Tests/CSharpCodeSnippetTest.cs
+++ b/src/CodeGeneration.Roslyn.Tests/CSharpCodeSnippetTest.cs
@@ -1,0 +1,59 @@
+﻿// Copyright (c) Andrew Arnott. All rights reserved.
+// Licensed under the MS-PL license. See LICENSE.txt file in the project root for full license information.
+
+namespace CodeGeneration.Roslyn.Tests
+{
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using Xunit;
+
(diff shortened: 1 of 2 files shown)


amis92:
Hi! It is a very interesting idea, however I don't think it has a place in this project. I' much more likely see this as an actual plugin project.

This project only delivers bare minimum framework to run your generators with.

Of course, if such a plugin would be written, we could feature it as a good starting point.

amis92:
@AArnott what do you think?

AArnott:
I agree.

PR #178, opened 2019-12-04
Use .NET Core SDK 3.1

### LINKED ISSUE #153: Upgrade build pipelines
* Build using dotnet build on all platforms (prerequisite: use only x-plat TFMs)
* Drop passing around build results (from Windows to others)
* Build Sample solutions using just-built nugets

### LINKED ISSUE #177: Move to .NET Core 3.1 SDK
Since there's a new LTS release of the SDK, we're moving to that.

SDK Announcement: https://github.com/dotnet/announcements/issues/142

amis92:
closes #177
closes #153


amis92:
So it turns out, out of windows, macos and ubuntu, only Ubuntu is path-case-sensitive.

Rulesets had wrong casing when they referenced another one in <Include> elements. And only on Ubuntu, things that were reported as Info or Hidden elsewhere, surfaced as Warnings.

Fixed by correcting reference casing.

amis92:
Who'd have thought. xD

PR #187, opened 2020-01-06
Make the generator cancellable

manne:
I did not spend much time where the `CancellationToken`s are observed and how to behave on the cancellation.
I am open for suggestions.

### DIFF
--- a/src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
+++ b/src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
@@ -84,7 +84,7 @@ namespace CodeGeneration.Roslyn.Engine
         /// <param name="progress">Optional handler of diagnostics provided by code generator.</param>
         /// <param name="cancellationToken">Cancellation token to interrupt async operations.</param>
         /// <returns>A <see cref="Task.CompletedTask"/>.</returns>
-        public async Task GenerateAsync(IProgress<Diagnostic> progress = null, CancellationToken cancellationToken = default)
+        public async Task GenerateAsync(IProgress<Diagnostic> progress, CancellationToken cancellationToken)
         {
             Verify.Operation(this.Compile != null, $"{nameof(Compile)} must be set first.");
             Verify.Operation(this.ReferencePath != null, $"{nameof(ReferencePath)} must be set first.");
@@ -104,13 +104,14 @@ namespace CodeGeneration.Roslyn.Engine
                 {
                     cancellationToken.ThrowIfCancellationRequested();
 
-                    string sourceHash = Convert.ToBase64String(hasher.ComputeHash(Encoding.UTF8.GetBytes(inputSyntaxTree.FilePath)), 0, 6).Replace('/', '-');
+                    string sourceHash = Convert.ToBase64String(hasher.ComputeHash(Encoding.UTF8.GetBytes(inputSyntaxTree.FilePath)), 0, 6).Replace(Path.AltDirectorySeparatorChar, '-').Replace(Path.DirectorySeparatorChar, '-');
                     Logger.Info($"File \"{inputSyntaxTree.FilePath}\" hashed to {sourceHash}");
-                    string outputFilePath = Path.Combine(this.IntermediateOutputDirectory, Path.GetFileNameWithoutExtension(inputSyntaxTree.FilePath) + $".{sourceHash}.generated.cs");
+                    string outputFilePath = Path.Combine(this.IntermediateOutputDirectory, Path.GetFileNameWithoutExtension(inputSyntaxTree.FilePath) + FormattableString.Invariant($".{sourceHash}.generated.cs"));
+                    var outputFile = new FileInfo(outputFilePath);
 
                     // Code generation is relatively fast, but it's not free.
                     // So skip files that haven't changed since we last generated them.
-                    DateTime outputLastModified = File.Exists(outputFilePath) ? File.GetLastWriteTime(outputFilePath) : DateTime.MinValue;
+                    DateTime outputLastModified = outputFile.Exists ? outputFile.LastWriteTime : DateTime.MinValue;
                     if (File.GetLastWriteTime(inputSyntaxTree.FilePath) > outputLastModified || assembliesLastModified > outputLastModified)
                     {
                         int retriesLeft = 3;
@@ -123,13 +124,13 @@ namespace CodeGeneration.Roslyn.Engine
                                     inputSyntaxTree,
                                     this.ProjectDirectory,
                                     this.LoadPlugin,
-                                    progress);
-
-                                var outputText = generatedSyntaxTree.GetText(cancellationToken);
-                                using (var outputFileStream = File.OpenWrite(outputFilePath))
+                                    progress,
+                                    cancellationToken);
+                                var outputText = await generatedSyntaxTree.GetTextAsync(cancellationToken);
+                                using (var outputFileStream = outputFile.OpenWrite())
                                 using (var outputWriter = new StreamWriter(outputFileStream))
                                 {
-                                    outputText.Write(outputWriter);
+                                    outputText.Write(outputWriter, cancellationToken);
 
                                     // Truncate any data that may be beyond this point if the file existed previously.
                                     outputWriter.Flush();
@@ -142,7 +143,7 @@ namespace CodeGeneration.Roslyn.Engine
                                     bool anyTypesGenerated = root.DescendantNodes().OfType<TypeDeclarationSyntax>().Any();
                                     if (!anyTypesGenerated)
                                     {
-                                        this.emptyGeneratedFiles.Add(outputFilePath);
+                                        this.emptyGeneratedFiles.Add(outputFile.FullName);
                                     }
                                 }
                                 break;
@@ -150,7 +151,7 @@ namespace CodeGeneration.Roslyn.Engine
                             catch (IOException ex) when (ex.HResult == ProcessCannotAccessFileHR && retriesLeft > 0)
                             {
                                 retriesLeft--;
-                                Task.Delay(200).Wait();
+                                await Task.Delay(200, cancellationToken);
                             }
                             catch (Exception ex)
                             {
@@ -162,7 +163,7 @@ namespace CodeGeneration.Roslyn.Engine
                         while (true);
                     }
 
-                    this.generatedFiles.Add(outputFilePath);
+                    this.generatedFiles.Add(outputFile.FullName);
                 }
             }
 
@@ -232,10 +233,7 @@ namespace CodeGeneration.Roslyn.Engine
 
             var location = inputSyntaxTree != null ? Location.Create(inputSyntaxTree, TextSpan.FromBounds(0, 0)) : Location.None;
 
-            var messageArgs = new object[]
-            {
-                ex,
-            };
+            var messageArgs = new object[] { ex };
 
             var reportDiagnostic = Diagnostic.Create(descriptor, location, messageArgs);
 
--- a/src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
+++ b/src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
@@ -84,7 +84,7 @@ namespace CodeGeneration.Roslyn.Engine
         /// <param name="progress">Optional handler of diagnostics provided by code generator.</param>
         /// <param name="cancellationToken">Cancellation token to interrupt async operations.</param>
         /// <returns>A <see cref="Task.CompletedTask"/>.</returns>
-        public async Task GenerateAsync(IProgress<Diagnostic> progress, CancellationToken cancellationToken)
+        public async Task GenerateAsync(IProgress<Diagnostic> progress = null, CancellationToken cancellationToken = default)
         {
             Verify.Operation(this.Compile != null, $"{nameof(Compile)} must be set first.");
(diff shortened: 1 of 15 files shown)


manne:
@amis92 I incorporated your feedback. Should I undo the unassociated changes in `CompilationGenerator` too?

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -84,7 +84,7 @@ public class CompilationGenerator
         /// <param name="progress">Optional handler of diagnostics provided by code generator.</param>
         /// <param name="cancellationToken">Cancellation token to interrupt async operations.</param>
         /// <returns>A <see cref="Task.CompletedTask"/>.</returns>
-        public async Task GenerateAsync(IProgress<Diagnostic> progress = null, CancellationToken cancellationToken = default)
+        public async Task GenerateAsync(IProgress<Diagnostic> progress, CancellationToken cancellationToken)
```
amis92: `CancellationToken` arguments are always last, and always optional. It's a good convention and we should keep it as such.

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -104,13 +104,14 @@ public async Task GenerateAsync(IProgress<Diagnostic> progress = null, Cancellat
                 {
                     cancellationToken.ThrowIfCancellationRequested();
 
-                    string sourceHash = Convert.ToBase64String(hasher.ComputeHash(Encoding.UTF8.GetBytes(inputSyntaxTree.FilePath)), 0, 6).Replace('/', '-');
+                    string sourceHash = Convert.ToBase64String(hasher.ComputeHash(Encoding.UTF8.GetBytes(inputSyntaxTree.FilePath)), 0, 6).Replace(Path.AltDirectorySeparatorChar, '-').Replace(Path.DirectorySeparatorChar, '-');
```
amis92: This is incorrect. It's specifically replacing slashes in the resulting hash string, it's not the path string.

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -104,13 +104,14 @@ public async Task GenerateAsync(IProgress<Diagnostic> progress = null, Cancellat
...
                     Logger.Info($"File \"{inputSyntaxTree.FilePath}\" hashed to {sourceHash}");
-                    string outputFilePath = Path.Combine(this.IntermediateOutputDirectory, Path.GetFileNameWithoutExtension(inputSyntaxTree.FilePath) + $".{sourceHash}.generated.cs");
+                    string outputFilePath = Path.Combine(this.IntermediateOutputDirectory, Path.GetFileNameWithoutExtension(inputSyntaxTree.FilePath) + FormattableString.Invariant($".{sourceHash}.generated.cs"));
```
amis92: As much as it's a good change, why is that included in this PR?

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -104,13 +104,14 @@ public async Task GenerateAsync(IProgress<Diagnostic> progress = null, Cancellat
...
+                    var outputFile = new FileInfo(outputFilePath);
```
amis92: I'm not exactly convinced this is necessary. An additional overhead, and doesn't make code much cleaner.

amis92: Also, again, it's a change introducing noise to this PR.

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -232,10 +233,7 @@ private static void ReportError(IProgress<Diagnostic> progress, string id, Synta
 
             var location = inputSyntaxTree != null ? Location.Create(inputSyntaxTree, TextSpan.FromBounds(0, 0)) : Location.None;
 
-            var messageArgs = new object[]
-            {
-                ex,
-            };
+            var messageArgs = new object[] { ex };
```
amis92: Again, a refactoring change.

src/CodeGeneration.Roslyn.Engine/DocumentTransform.cs
```diff
@@ -42,13 +42,15 @@ public static class DocumentTransform
...
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
         /// <returns>A task whose result is the generated document.</returns>
         public static async Task<SyntaxTree> TransformAsync(
             CSharpCompilation compilation,
             SyntaxTree inputDocument,
             string projectDirectory,
             Func<AssemblyName, Assembly> assemblyLoader,
-            IProgress<Diagnostic> progress)
+            IProgress<Diagnostic> progress,
+            CancellationToken cancellationToken)
```
amis92: `CancellationToken` arguments are always last, and always optional. It's a good convention and we should keep it as such.

src/CodeGeneration.Roslyn.Tests/Helpers/CompilationTestsBase.cs
```diff
@@ -78,7 +79,7 @@ protected static async Task<SyntaxTree> GenerateAsync(string source)
         var diagnostics = compilation.GetDiagnostics();
         Assert.Empty(diagnostics.Where(x => x.Severity >= DiagnosticSeverity.Warning));
         var progress = new Progress<Diagnostic>();
-        var result = await DocumentTransform.TransformAsync(compilation, tree, null, Assembly.Load, progress);
+        var result = await DocumentTransform.TransformAsync(compilation, tree, null, Assembly.Load, progress, CancellationToken.None);
```
amis92: Won't be necessary if the parameter is kept optional.

src/CodeGeneration.Roslyn.Tool/Program.cs
```diff
@@ -14,6 +16,24 @@ namespace CodeGeneration.Roslyn.Generate
     internal static class Program
     {
         private static async Task<int> Main(string[] args)
+        {
+            using (var cancellationTokenSource = new CancellationTokenSource())
```
amis92: I'd suggest to initialize this `cts` besides `progress` and dispose in the `finally` of the existing `try`.

@AArnott what's your opinion?

manne:
@amis92  I hope to be able to work on the requested changes during the weekend

src/CodeGeneration.Roslyn.Tool/Program.cs
manne: My intention is to separate the cancellation aspect from the rest (here `Core` method)

AArnott:
Thanks, @amis92. I'll take a look.

src/CodeGeneration.Roslyn.Engine/DocumentTransform.cs
```diff
@@ -42,13 +42,15 @@ public static class DocumentTransform
...
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
         /// <returns>A task whose result is the generated document.</returns>
         public static async Task<SyntaxTree> TransformAsync(
             CSharpCompilation compilation,
             SyntaxTree inputDocument,
             string projectDirectory,
             Func<AssemblyName, Assembly> assemblyLoader,
-            IProgress<Diagnostic> progress)
+            IProgress<Diagnostic> progress,
+            CancellationToken cancellationToken = default)
```
AArnott: Is there any good reason the caller shouldn't provide one of these? So long as we're making breaking API changes as this does, should we mandate that they provide one?

src/CodeGeneration.Roslyn.Tool/Program.cs
AArnott: Either way is fine. I'd vote to go with what @manne has already done. This is *very* similar to [my prescriptive pattern](https://gist.github.com/AArnott/daa2e450fb5af165f5ca2aff42a4b5cc).

src/CodeGeneration.Roslyn.Engine/DocumentTransform.cs
amis92: Well, to be frank I don't consider the `Engine` part to be a "public" API. But it is, obviously. So maybe we could just make it not be public?

On the actual issue: I think both `CancellationToken` and `IProgress` should still be pay-for-play. In most cases I've seen, they are optional. Or rather, I've not seen a case where they're required. But it's not a strong opinion, since as I've mentioned above, I'd not expect one to use the `Engine` package alone, and not consider it a supported scenario.

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -150,7 +150,7 @@ public async Task GenerateAsync(IProgress<Diagnostic> progress = null, Cancellat
                             catch (IOException ex) when (ex.HResult == ProcessCannotAccessFileHR && retriesLeft > 0)
                             {
                                 retriesLeft--;
-                                Task.Delay(200).Wait();
+                                await Task.Delay(200, cancellationToken);
                             }
                             catch (Exception ex)
```
amis92: Should we maybe handle [`OperationCancelledException`](https://docs.microsoft.com/en-us/dotnet/api/system.operationcanceledexception?view=netframework-4.8) here as a special case then?

If the operation got cancelled, I'd not expect generation to continue on all the other files in queue.

manne: Proposal A: in case of a cancellation the `while` loop could be exited, then in the `foreach` the flow would be cancelled. Additionally after the `foreach` an `cancellationToken.ThrowIfCancellationRequested()` should be added.

Proposal B: add an filter `catch(Exception ex) when(!(ex is OperationCanceledException))`

After writing these proposals I prefer proposal B

src/CodeGeneration.Roslyn.Engine/DocumentTransform.cs
manne: I personally prefer that the cancellationtoken is mandatory. For me, cancellation is a crucial aspect of asynchronous behavior.
I am OK leaving at it is (PR), and/or making the `progress` optional, too.

amis92: Well, let's try a different approach. Would you agree that required arguments are those without which the method cannot successfully execute?

And would you agree then that others could be called optional?

And finally, that the methods we discuss can execute *without* either CTS or Progress? :)

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
amis92: I think that's the simplest and most clear solution. And additionally, doesn't interfere with stacktraces, so an additional bonus.

src/CodeGeneration.Roslyn.Engine/DocumentTransform.cs
AArnott: Supporting `null` for `IProgress<T>` and `default(CancellationToken)` are different questions than whether we make the parameters themselves optional by providing default values.

I fully support any method working with null/None values for both of these parameter types.

However making a parameter *optional* makes it very easy for callers to neglect to provide arguments for them even when they are available. Making `cancellationToken` optional for example means that our (one?) caller might forget to pass one in, even though it might very well have a `cancellationToken` that _could_ be passed in!

So in general when designing APIs, I only make parameters optional if it's reasonable to assume that many callers will *legitimately* not want to pass in an argument. But if most/all of my callers should have a value for a parameter, I make it mandatory so they don't forget to provide it.

manne: I am lost in the discussion ;) How should the parameters be?

A) progress = null, cancellationToken = default
B) progress,  cancellationToken = default
C) progress,  cancellationToken

I prefer C :)

amis92: @AArnott does "potential callers" include (unit) tests? Because in tests, it's very much legitimate to not want to provide those, and specifying those as "default" doesn't convey anything, it's just additional noise - that's how I see it.

That aside, I do understand the reasoning you provided. I've also looked at [.NET Guide for TAP](https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap#cancellation-optional) and indeed there's no implied requirement for those parameters to be optional in the sense of syntax declarations, so I'm also going to *agree on making both parameters non-optional* syntax-wise.

manne:
@AArnott, @amis92 I see that both of you approved this PR. From my point of view this PR can be merged. I am not allowed to to this. Should I change something?

src/CodeGeneration.Roslyn.Engine/DocumentTransform.cs
AArnott: When it comes to public APIs, I do not optimize for tests. I optimize for expected actual users. That said, if my tests don't represent my typical users then there's something likely wrong with how I'm testing.

But most of my unit tests (not necessarily in this repo) *do* have `CancellationToken` available in order to help them gracefully timeout if they run too long. 

At the end of the day, I'm not too passionate either way on this question, both because the users of this API are few and because I think if a few of those callers do it wrong and omit the cancellation token, I'm not going to lose any sleep over it.

But yes, I'm suggesting option C.

Issue #183, opened 2019-12-13
Support C#8 NRT (Nullable Reference Types) in generated code


amis92:
It was brought to my attention that generating code in NRT-enabled projects raises warnings.

Reference: https://github.com/amis92/RecordGenerator/issues/110

---

*From the issue linked above:*

Looks like the dotnet team decided auto-generated code needs to explicitly opt-in into nullable reference types:

```c#
public partial class TestClass
{
    public string? SomeTestProperty { get; }
}
```

In a project with
```xml
<LangVersion>8.0</LangVersion>
<Nullable>enable</Nullable>
```
this generates without issues and is actually usable, but the following compiler warning is listed for the generated file. One warning per line that uses `?` on a reference type.

> Warning	CS8669	The annotation for nullable reference types should only be used in code within a '#nullable' annotations context. Auto-generated code requires an explicit '#nullable' directive in source.

---

What can/should we do about it? We'll definitely need to get aware of the Nullable project property. What else can we do to simplify/enable nullable generated code?

AArnott:
To the extent that we can generate nullable ref annotated code regardless of the rest of the project, I think we should. But emitting the `#nullable enable` directive shouldn't be done unless the actual generator is participating or else we're emitting nullable annotations even if no `?` is added anywhere. 

So IMO this is something each individual generator should do if it wants to. A generator that does *not* want to participate would need to shed the `?` annotations as appropriate wherever it copies types around from user code.

amis92:
So let's consider a couple of scenarios for C#8 projects:

1. Project is nullable-oblivious - no problem, same as pre-8.0
2. Project is nullable-enabled, generator is oblivious:
  - case as the OP's report linked, nullable declarations "leak" to the generated code.
  - fix: ignore the specific CS8669 warning
  - can we do better? Generator could attempt to "strip" the annotations but that requires generator to be updated. CGR could "wrap" any code from oblivious generators in a pragma that disables that warnings, making using "older" generators painless?
3. Project is nullable-enabled, generator attempts to as well:
  - right now, generator would have to read project file on it's own - we don't provide information whether project uses Nullable or what langversion it targets.
  - assuming the generator knows these, it can "switch" to produce nullable-conscious code and wrap it in `#nullable enable` directives to become independent of other generators
  - framework can provide a concept of generator manifest that allows generators to declare nullable-compatibility and automatically wrap generated code in appropriate directives
[...]

PR #190, opened 2020-01-28
Add example for a generator respecting nullability

manne:
This PR adds a generator respecting the nullability context of the class.

For the class `NullableEnabled` where `#nullable enable` it generates a method `string? DoSomething(string?)`. In the generated code nullable is enabled and restored properly.

For the class `NullableDisabled` where `#nullable disable` it generates a method `string DoSomething(string)`.

can solve issue #183 . I did not read the whole conversation properly 🤨

### DIFF
--- a/src/CodeGeneration.Roslyn.Tests/CodeGeneration.Roslyn.Tests.csproj
+++ b/src/CodeGeneration.Roslyn.Tests/CodeGeneration.Roslyn.Tests.csproj
@@ -7,7 +7,6 @@
     <GenerateCodeFromAttributesToolPathOverride>$(OutputPath)..\..\CodeGeneration.Roslyn.Tool\$(Configuration)\netcoreapp2.1\dotnet-codegen.dll</GenerateCodeFromAttributesToolPathOverride>
     <!-- Ignore CGR1002 because it's expected to be raised by legacy Amadevus.RecordGenerator -->
     <MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);CGR1002</MSBuildWarningsAsMessages>
-    <LangVersion>8.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/CodeGeneration.Roslyn.Tests/CodeGeneration.Roslyn.Tests.csproj
+++ b/src/CodeGeneration.Roslyn.Tests/CodeGeneration.Roslyn.Tests.csproj
@@ -7,6 +7,7 @@
     <GenerateCodeFromAttributesToolPathOverride>$(OutputPath)..\..\CodeGeneration.Roslyn.Tool\$(Configuration)\netcoreapp2.1\dotnet-codegen.dll</GenerateCodeFromAttributesToolPathOverride>
     <!-- Ignore CGR1002 because it's expected to be raised by legacy Amadevus.RecordGenerator -->
     <MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);CGR1002</MSBuildWarningsAsMessages>
+    <LangVersion>8.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


src/CodeGeneration.Roslyn.Tests/OtherFolder/NullableGeneratorTests.cs
```diff
@@ -0,0 +1,48 @@
+﻿namespace CodeGeneration.Roslyn.Tests.OtherFolder
+{
+    using System;
+    using System.Diagnostics;
+    using CodeGeneration.Roslyn.Tests.Generators;
+    using Xunit;
+
+    public class NullableGeneratorTests
+    {
+#nullable enable
```
amis92: I don't know much about conventions in this regard, but I think when whole file isn't `#nullable enable`d, every explicit switch should be paired with `#nullable restore` so that in future when we migrate this project to full nullability, everything works correctly.

Also, I feel like for Facts, the directives should be inside method body.

src/CodeGeneration.Roslyn.Tests/CodeGeneration.Roslyn.Tests.csproj
```diff
@@ -7,6 +7,7 @@
     <GenerateCodeFromAttributesToolPathOverride>$(OutputPath)..\..\CodeGeneration.Roslyn.Tool\$(Configuration)\netcoreapp2.1\dotnet-codegen.dll</GenerateCodeFromAttributesToolPathOverride>
     <!-- Ignore CGR1002 because it's expected to be raised by legacy Amadevus.RecordGenerator -->
     <MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);CGR1002</MSBuildWarningsAsMessages>
+    <LangVersion>8.0</LangVersion>
```
amis92: Can we maybe move this up to `Directory.Build.props`? I think it shouldn't add any warnings.

amis92 (approved):

amis92:
So, this example is fine and a good addition. It doesn't fix #183, however.

That would require passing around both `LangVersion` and `Nullable` msbuild properties, starting in `.targets`, through `dotnet-codegen` up to the creation of `Compilation`.

src/CodeGeneration.Roslyn.Tests/CodeGeneration.Roslyn.Tests.csproj
manne: > Can we maybe move this up to Directory.Build.props? I think it shouldn't add any warnings.
I'll do. It has 0 warnings afterwards.

manne: The change adds some messages, stating to use some C# 8.0 features. `IDE0063 'using' statement can be simplified`, ...

manne:
The following code/file is generated

```CSharp
// ------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by a tool.
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
// ------------------------------------------------------------------------------

namespace CodeGeneration.Roslyn.Tests.OtherFolder
{
    using System;
    using System.Diagnostics;
    using CodeGeneration.Roslyn.Tests.Generators;
    using Xunit;

    public partial class NullableEnabled
    {
#nullable enable
        public string? DoSomething(string? value)
        {
            return value;
        }
#nullable restore

#nullable enable
        public string? DoSomethingStrict(string value)
        {
            return value;
        }
#nullable restore
    }
}

namespace CodeGeneration.Roslyn.Tests.OtherFolder
{
    using System;
    using System.Diagnostics;
    using CodeGeneration.Roslyn.Tests.Generators;
    using Xunit;

    public partial class NullableDisabled
    {
        public string DoSomething(string value)
        {
            return value;
        }

        public string DoSomethingStrict(string value)
        {
            return value;
        }
    }
}
```

amis92:
I think that since we need to check that the warning *is* generated (if we don't specify `#pragma warning disable`), we should write tests more like `DocumentTransformTests` are, with the use of `CompilationTestsBase`.

So, inputs and outputs are just strings that contain source code, and we could validate that the resulting compilation has no errors/warnings by checking diagnostics.

It may be probably done in another PR, though.

AArnott:
This project is getting archived in favor of Source Generators. See #229

PR #198, opened 2020-03-25
[WIP] Refactor: Tool now self-contained, BuildTime merged into it

amis92:
This is a major refactor.

It's triggered by a disaster of what 0.7.5-alpha turned out to be.

Major features:
* `dotnet-codegen` package is now `CodeGeneration.Roslyn.Tool`, as per project name.
  This was done to prevent confusion with previous CLI tool, which it now is not at all,
  in terms of either DotNetCliToolReference-style tools and global/local .NET CLI tools.
* `Tool` is now a `DevelopmentDependency=true` normal Dependency package (as in,
  suitable for `PackageReference` usage. Via various properties, it's setup so that it contains
  a framework-dependent executable (.dll) with all dependencies included. Using `build/CG.R.Tool.props` file, the path to the executable is set in referencing projects.
* `BuildTime` package's props/targets were merged into the `Tool` package because a separate package is no longer necessary, and as such necessary consumer dependencies are trimmed down
to just `CodeGeneration.Roslyn.Tool` package and a generator plugin of choice.
* Samples added: samples that publish generator packages were added, including one that has an external dependency (Newtonsoft.Json). There's also a build script that uses locally built CG.R packages; that script is also called during GitHub Actions CI workflow.

TODO:
- [x] Update Sdk package as per sample packable generators' csproj files
- [x] Update samples to use Sdk props/targets directly (not via Sdk element).
- [x] Update README
- [x] Update CHANGELOG
- [x] Note down migration steps

To use Sdk package as an Sdk element, the Sdk will need to be published to nuget.org, because MSBuildSdkResolver doesn't work well with local feeds.


amis92:
@AArnott as seen in https://github.com/AArnott/CodeGeneration.Roslyn/runs/534972275#step:9:12

Nerdbank.GitVersioning v3.1 introduces ExcludeFromCodeCoverage on ThisAssembly,
and the attribute is not available in net20 target from Attributes pkg.

I had to downgrade. I know net20 is not supported, but maybe you could consider doing an exception for it like you did for netstandard1.x in https://github.com/dotnet/Nerdbank.GitVersioning/pull/453 ?

amis92:
I've updated the wiki for the newest changes: https://github.com/AArnott/CodeGeneration.Roslyn/wiki/Migrations

amis92:
@AArnott macos env in AzurePipelines times out since some time.

AArnott:
@amis92 I see you already fixed the macOS problem. :)

PR #208, opened 2020-04-03
refactor: Remove .Tests.RunBuild project

amis92:
It doesn't seem to be used for anything,
and debugging can be done via more "Unit test" tests


AArnott:
It's your call, @ami92. But the useful of this project is (IIRC) that you can "F5" and actually run a build that actually invokes the msbuild task in the regular way. There were some bugs that I couldn't write effective unit tests for, perhaps. Or perhaps it was just easier for me to step through debugging using this project.

PR #210, opened 2020-04-04
Give Generators Access to MSBuild Properties

talenfisher:
I have a use-case where I need to access MSBuild properties of the project I am generating code for, including TargetFrameworkVersion and OutDir, plus some custom ones.  This implementation allows for that, using an opt-in approach:

1.  Add the name of the build property to the PluginRequestedProperty item list in the generator's .csproj file.
2. Access the value via TransformationContext.BuildProperties["PropertyName"]


amis92:
Hi, thanks for the PR. This is a larger topic, so I'll come back to this PR in vNext (not 0.7).

manne:
@talenfisher works like a charm!

src/CodeGeneration.Roslyn.Tool/CommandLine/Enumerable.cs
```diff
@@ -64,6 +64,11 @@ public static T[] ToArray<T>(this IEnumerable<T> source)
             return System.Linq.Enumerable.ToArray(source);
         }
 
+        public static Dictionary<TKey, TElement> ToDictionary<T, TKey, TElement>(this IEnumerable<T> source, Func<T, TKey> keySelector, Func<T, TElement> elementSelector)
```
amis92, line 67: It's not clear to me why is this needed. Looks like it just duplicates/wraps the same signature as Linq.

manne:
With this functionality I can access the `Nullable` state to generate correct nullable code

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
```diff
@@ -54,7 +54,10 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
-    <CodeGenerationRoslynPlugin Include="%24(MSBuildThisFileDirectory)../$(PluginPackagePath)/$(TargetFileName)" />
+    <CodeGenerationRoslynPlugin Include="%24(MSBuildThisFileDirectory)../$(PluginPackagePath)/$(TargetFileName)" />  
+  </ItemGroup>
+  <ItemGroup>
+    @(GeneratorBuildPropertyNames->'<GeneratorBuildPropertyNames Include="%(Identity)" />')
```
amis92, line 60: Let's format this nicely. it can be in the same ItemGroup as CodeGenerationRoslynPlugin.

I'd even question whether it shouldn't/couldn't be a Property.

Either way, for a well-formatting transform see https://github.com/AArnott/CodeGeneration.Roslyn/blob/22671d2d76ea52c699ec3adad6b13b7cb1ada1e3/src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets#L64

amis92:
@AArnott I'd like you to approve this before we work on it further. My opinion is that it's a nice balance between what Uno.SourceGeneration did (expose whole msbuild context) and what we want (semi-clear Compilation). We already expose some selected properties, like ProjectDirectory or DefineConstants. With this, we could just by default do `<BuildProperties Include="MSBuildProjectDirectory" />` and have that covered.

---

@talenfisher 
If we get approval, I'd have some change requests.
- MSBuild item groups should have singular names, e.g. BuildProperty.
- I'd prefer the itemgroup name to start with `CodeGenerationRoslyn`; my personal suggestion would be `CodeGenerationRoslynRequestedProperty` since it's a request for inclusion.
- we should deduplicate the list before processing it into key-value pairs (`->Distinct()`)

---
@manne Nullable will be handled separately. I'd expect handling nullability to ship before this. This is a major extension point. We'll analyze whether it's a good idea to add.

src/CodeGeneration.Roslyn.Tool/CommandLine/Enumerable.cs
talenfisher: This was needed to get [this sanitization method](https://github.com/AArnott/CodeGeneration.Roslyn/pull/210/files/9bfa30c5a7f343b7be84976a72399ff8e8de69be#diff-9070cd2b914605c42c39f67b4431b352R140-R143) to work.  I tried adding a using for System.Linq in that Program.cs file, but got a name clash between linq enumerables and codegeneration.roslyn.tool.commandline enumerables.  I could use the fully qualified linq name in the Program.cs file, but this seemed cleaner and more consistent.

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
talenfisher: I'll experiment with converting this to a property today and make the requested de-duplication, name, and formatting changes.  Thanks for your review!

AArnott:
> MSBuild item groups should have singular names

That's fine, but just a nit: *This* is an MSBuild item group:

```xml
<ItemGroup>
</ItemGroup>
```

*These* are items:

```xml
  <Something Include="a.txt" />
  <Else Include="b.txt" />
```

*These* are item *types*: `Something`, `Else`.
And finally, this is an item *list*: `@(Else)`.

Just a pet peeve of mine that people use "item group" instead of "item type". :)

README.md
```diff
@@ -457,6 +458,28 @@ as it's dependant packages.
...
 
+### Accesss MSBuild Properties
+
+You may access MSBuild property values of the project being generated for, by first adding the property
+name to the `GeneratorBuildPropertyNames` item group.  For example, if you want to access the TargetFramework build
+property, you would do the following in your generator's .csproj file:
+
+```xml
+<ItemGroup>
+  <GeneratorBuildPropertyNames Include="TargetFramework">
```
AArnott, line 469: This is invalid xml. Replace the last character with `/>` to make the xml element self-closing.

src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
```diff
@@ -24,6 +24,10 @@
         Include="@(Compile)"
         Condition=" '%(Compile.Generator)' == 'MSBuild:GenerateCodeFromAttributes' " />
       <_CodeGenToolInputs_DefineConstants Include="$(DefineConstants)" />
+      <GeneratorBuildProperties Include="@(GeneratorBuildPropertyNames)">
+        <Key>%(Identity)</Key>
+        <Value>$(%(Identity))</Value>
+      </GeneratorBuildProperties>
       <!-- Map GeneratorAssemblySearchPaths to actual DLLs until support is removed -->
```
AArnott, line 31: I'd never thought of this syntax to pass arbitrary properties to an msbuild task. It's very clever.

manne:
the functionality of this PR would fix the warnings of #190

src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
amis92: About that, I'm wondering whether we shouldn't escape `;` somehow in the Value. If one wanted to pass `TargetFrameworks`, it'd definitely break response file construction and parsing.

AArnott: Good thought. I wouldn't escape it *here* though, since in general an MSBuild item can have arbitrary values for metadata. But when in your msbuild task you create the response file, you should apply whatever escaping is necessary so that the next step will properly parse it.

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
talenfisher: I am able to make this a property, however, I need to create an additional item group in the tool targets to get this to work, for example:

```xml
<_CodeGenerationRoslynRequestedProperty Include="$(CodeGenerationRoslynRequestedProperties)" />
<_CodeGenerationRoslynRequestedPropertyResolved Include="@(_CodeGenerationRoslynRequestedProperty)">
  <Key>%(Identity)</Key>
  <Value>$(%(Identity))</Value>
</_CodeGenerationRoslynRequestedPropertyResolved>
```

Doing the following results in %(Identity) evaluating to a blank string:
```xml
<_CodeGenerationRoslynRequestedPropertyResolved Include="$(CodeGenerationRoslynRequestedProperties)">
  <Key>%(Identity)</Key>
  <Value>$(%(Identity))</Value>
</_CodeGenerationRoslynRequestedPropertyResolved>
```

So, we'll need it in item list form prior to building the 'resolved' item list, that is with key and value metadata.  From the user's perspective, I can see why this should maybe be a property instead though, ie less clutter...  what are your thoughts?

amis92: @AArnott what do you think? For a simple list of strings, Property seems more... appropriate.
We'll create new item type anyway to "parse" it and set value metadata in the target, so that wouldn't change.

README.md
talenfisher: Oops, good catch.  Replaced this with valid xml in 6deb257

src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
talenfisher: Now escaping semicolons in 6deb257

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
talenfisher: Made the switch to using a property in 6deb257.  I think it definitely looks cleaner.  I'll change it back to an item if that turns out to be preferable.

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
```diff
@@ -56,9 +56,12 @@
   <ItemGroup>
     <CodeGenerationRoslynPlugin Include="%24(MSBuildThisFileDirectory)../$(PluginPackagePath)/$(TargetFileName)" />  
   </ItemGroup>
-  <ItemGroup>
-    @(GeneratorBuildPropertyNames->'<GeneratorBuildPropertyNames Include="%(Identity)" />')
-  </ItemGroup>
+  <PropertyGroup>
+    <CodeGenerationRoslynRequestedProperties>
+      %24(CodeGenerationRoslynRequestedProperties);
+      $(CodeGenerationRoslynRequestedProperties)
```
amis92, line 71: I think the property used in Plugin project should be named differently, e.g. `PluginPackageRequestedProperties` - as in, this should be the Property used in Plugin csproj to request these properties be added in the package `targets`.

In the target that creates the package targets (the line commented on), we add `PluginPackageRequestedProperties` to `CodeGenerationRoslynRequestedProperties`.

The distinction would be necessary if for example the Plugin project itself used the Tool and some Plugins to generate some internal code. In that case, current code would add RequestedProperties added by *plugins used by* our Plugin.

It's a lot like inception ;) If it's not clear, I can offer an example.

src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
amis92: I think @AArnott is right. However I don't know how exactly to do that. Possibly, we need to escape it directly in `_CodeGenToolResponseFileLines` below. At that point we actually use the semicolons to separate lines in response file via including the property into an item type `_CodeGenToolResponseFileContent`.

amis92: Oooh also, I'm not sure - do we need to Trim `Key` of whitespace? I think we might need to.

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
talenfisher: Think I get it haha, please check 56a3641 and let me know if I did it right.

src/CodeGeneration.Roslyn.Tool/CommandLine/Enumerable.cs
amis92: Yeah we should probably get rid of that commandline stuff somehow :/

src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
talenfisher: Originally, I had the transformations all in a single line in the part of the file where the response file lines are created, like this:

@(_CodeGenerationRoslynResolvedProperty->'%(Key)=%(Value)'->Replace(';', '%3B')->'--buildProperty;%(Identity)')

...but I felt that might be a little confusing, so created the 'escaped' item list.  Let me know if there is a better way.  I am also not sure about the necessity of trimming the key, but it seems like a good cautionary measure, so I'll add it in the next commit

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
amis92: LGTM - I'd prefer PluginRequestedProperties than PackageRequestedProperties if you want it shorter, because `Package*` prefix is mostly used by nuget pack targets - I'd avoid possibility of clash in future.

talenfisher: Ah, ok! Renamed to PluginRequestedProperties in 982c210

src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
talenfisher: Now trimming keys in 982c210

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
AArnott: 😢 MSBuild properties should store *scalars*. Use items for *vectors*.

The fact that some msbuild properties store delimited lists is a shame, and is probably a carryover from earlier versions of MSBuild where you couldn't remove items from an msbuild item list in the evaluation model. But that deficiency has since been resolved. All lists should be stored in item lists rather than delimited properties, IMO.
So to be clear, I prefered the `GeneratorBuildPropertyName` item approach.

src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
AArnott: > do we need to Trim Key of whitespace? I think we might need to.

You only need to if you're abusing properties for what items should be used for. :)

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
talenfisher: That makes sense.  Then, would everyone be ok with switching back to the items approach, but kept the current naming (with the exception of converting plural item names to singular)?

src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
amis92: So, if we add property content (possibly value like `   a ;   b; ;`) into an item, will it trim and split into a nice
- `a`
- `b`
list?

If it would, we're safe then since we use item for metadata.

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
amis92: Let's go with items then. Let's be *good practicians*. ;)

src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
amis92: I think since we're going with items, trimming shouldn't be necessary and escaping Value should happen at the response-file line creation.

talenfisher: Ok, removed trimming in e508e37 and moved escaping down to response file line creation in f9cda4b

AArnott: >  if we add property content (possibly value like a ; b; ;) into an item

Property concatenation patterns tend to produce lots of whitespace. One reason why they're *not* good at lists. So ya, if you start with them, you have to clean them up. But if you don't use them at all, the whitespace doesn't tend to happen.

src/CodeGeneration.Roslyn.Plugin.Sdk/build/BuildPluginPackage.targets
talenfisher: Got everything switched over to items in the last two commits - let me know if anything else should be changed, or if I should go ahead and squash.  Thanks again for the reviews!

amis92: I'll squash-merge the PR anyway, so no reason to force-push here.

talenfisher:
@amis92 I am thinking about renaming them to RequestedProperties in the transformation context, but I'm not sure that would make sense elsewhere (the tool + engine did not request them, rather they simply passed them down).  Would it be okay to leave them named as-is in the engine + tool?

talenfisher:
updates 4/13 per @amis92 suggestions:
- added changelog entry to unreleased
- added BuildPropsGenerator and BuildPropsConsumer samples, added link to these in readme
- updated doc comments in transformation context to be more concise

I think the doc comments make it clear enough now that these are not a 'full' dictionary of build properties, so I've left them named BuildProperties for now.  I'm not against renaming them though, if that is everyone's preference

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+* Support for accessing arbitrary MSBuild properties, see [Readme section](#access-msbuild-properties) ([#210])
```
amis92, line 10: I think the link should be `[…](README.md#access-msbuild-properties)`

talenfisher: woops - fixed 814b691

PR #211, opened 2020-04-04
Include custom props/targets files in packages

talenfisher:
A downside of the new SDK system introduced in 0.7 is that I can no longer (easily) add my own package props/targets.  The current workaround is to do something like:

1. Add a target that runs before pack to rename the build/packageId.{props|targets} files to something else and import them or
2. Rewrite the build/packageId.{props|targets} files to include your own

I would personally prefer to not do either of those, so this PR implements a way to configure the SDK to include them when it creates the build/packageId.{props|targets} files

TODO: 
- [ ] Update readme

### DIFF
--- a/samples/PackagedGenerator/PackagedGenerator.csproj
+++ b/samples/PackagedGenerator/PackagedGenerator.csproj
@@ -7,6 +7,11 @@
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
   </PropertyGroup>
 
+  <ItemGroup>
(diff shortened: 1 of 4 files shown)


amis92:
It's a good point that customization should be allowed. However, I don't like the approach, it's introducing unnecessary custom elements (like the ItemGroups CodeGenerationRoslynPackage...Files).

We can reuse None/Content/TfmSpecificPackageFile groups, and import items that have PackagePath="build*" for example.

I'll try to work on that.

amis92:
Tracking in #212 now.

PR #224, opened 2020-04-10
Pin .NET SDK version

manne:
I had difficulties locally to build, because I have installed a .NET SDK version of `5.0.100-alpha.1.20069.4`.  With this change `dotnet` will use the version from the `global.json` file on any environment.
I have the version number from a [build](https://github.com/AArnott/CodeGeneration.Roslyn/runs/570474519?check_suite_focus=true#step:4:8).


amis92:
But, we do have the version pinned: https://github.com/AArnott/CodeGeneration.Roslyn/blob/master/src/global.json

Also, we have a document that says what you need: https://github.com/AArnott/CodeGeneration.Roslyn/blob/master/CONTRIBUTING.md

So, if you want to move the file to the root, sure. But also you'll need to find and update all references to it (e.g. that Contributing doc, and maybe some pipelines that read/use the file? not sure).

manne:
>But, we do have the version pinned: https://github.com/AArnott/CodeGeneration.Roslyn/blob/master/src/global.json
> Also, we have a document that says what you need: https://github.com/AArnott/CodeGeneration.Roslyn/blob/master/CONTRIBUTING.md

😣 I have forgotten the content of the contributing file.

>So, if you want to move the file to the root, sure. But also you'll need to find and update all references to it (e.g. that Contributing doc, and maybe some pipelines that read/use the file? not sure).

Actually I added the `global.json` to the root, because I could not build the samples via the `samples\build.ps1` script.

@amis92 Is it OK to "move" the `global.json` file to the root? If so I will
- adapt the files needed for GitHub action and Azure Pipelines.
- change to content of `global.json` to the content of `src/global.json`

AArnott:
Yes, global.json should be in the root of any repo. This is because dotnet.exe gets the SDK from the _current directory or above_ rather than the directory of what it's building. So if for example the global.json file is under the `src` directory, and you `dotnet build src` from the root of the repo, you're _not_ using the SDK version that the global.json file says to use. 

The file should *not* also appear somewhere else in the repo as that simply replicates content that now must be maintained. 

But as @amis92 said, let's update references to it in our docs in the process of moving it.

amis92:
If you'd be so nice to re-add that file in the `.sln` as well with the correct path, that'd be awesome.

manne:
> If you'd be so nice to re-add that file in the `.sln` as well with the correct path, that'd be awesome.

@amis92 which file should I add to the solution?

amis92:
Well, global.json, since it was moved. It's added in Solution Files folder, but now it won't be found since it points to the old path.

amis92:
It's not really important here, but it seems your git settings don't have email linked with GitHub account - the commits aren't correctly marked as "yours".

Issue #184, opened 2019-12-13
<Nullable>enable</Nullable> this project


amis92:
https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/upgrade-to-nullable-references

PR #225, opened 2020-04-10
Nullable for the solution

manne:
- Implies usage of C# 8.0, instead of 7.3


manne:
see #184

manne:
I did not make the command line functionality nullable aware, as it has a "huge" complexity and some functionality which is not being used.
Is the source of command line functionality this repo [dotnet/command-line-api](https://github.com/dotnet/command-line-api)? 
Unfortunately that code base is not nullable aware.

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -74,23 +92,18 @@ public class CompilationGenerator
...
         /// <param name="progress">Optional handler of diagnostics provided by code generator.</param>
         /// <param name="cancellationToken">Cancellation token to interrupt async operations.</param>
         /// <returns>A <see cref="Task.CompletedTask"/>.</returns>
-        public async Task GenerateAsync(IProgress<Diagnostic> progress = null, CancellationToken cancellationToken = default)
+        public async Task GenerateAsync(IProgress<Diagnostic> progress, CancellationToken cancellationToken = default)
         {
-            Verify.Operation(this.Compile != null, $"{nameof(Compile)} must be set first.");
-            Verify.Operation(this.ReferencePath != null, $"{nameof(ReferencePath)} must be set first.");
-            Verify.Operation(this.PluginPaths != null, $"{nameof(PluginPaths)} must be set first.");
-            Verify.Operation(this.IntermediateOutputDirectory != null, $"{nameof(IntermediateOutputDirectory)} must be set first.");
```
amis92, line 92: Public API behavior (null validation) shouldn't be removed. NRTs don't change that - they don't actually prevent null from being passed.

Then again, this whole class had it's shape changed in this PR; @AArnott what's your PoV here?

manne: > Public API behavior (null validation) shouldn't be removed. NRTs don't change that - they don't actually prevent null from being passed.

That's true. I will add them

manne:
> Changing API shape (CompilationGenerator) also doesn't quite feel like it belongs in here. But that I'll leave up to @AArnott.

I have no problem to change it back,

> I think, if this PR is true to it's name, <Nullable> should be enabled in Directory.Build.props and then in projects that weren't/couldn't be migrated, disable it. Currently that'll be only Tests and Tool it seems?

I activated `nullable` globally. I found some interesting situations 🙂

manne:
To get rid of the build error

- we could update the nuget packge `StyleCop.Analyzers` to `1.2.0-beta.164`. see dotnetanalyzers/stylecopanalyzers#3098 or
- suppress the warning or 
- disable `nullable` for the corresponding project

src/CodeGeneration.Roslyn.Tests/Helpers/CompilationTestsBase.cs
```diff
@@ -21,7 +21,9 @@ public abstract class CompilationTestsBase
     static CompilationTestsBase()
     {
         // this "core assemblies hack" is from https://stackoverflow.com/a/47196516/4418060
+#pragma warning disable SA1009 // Closing parenthesis should be spaced correctly
```
manne, line 24: I temporarily disabled the warning to get the build running

AArnott: We should fix it by updating to stylecop.analyzers 1.2.0-beta.164

AArnott:
Once NRT is activated, I omit/remove null checks on all _internal_ methods.
But as @amis92 says, I keep null checks present on public entrypoints.

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -34,29 +33,48 @@ public class CompilationGenerator
...
         /// </summary>
-        public IReadOnlyList<string> Compile { get; set; }
+        /// <param name="projectDirectory">The list of paths of files to be compiled.</param>
+        /// <param name="compile">The directory with the project file.</param>
+        /// <param name="referencePath">The list of paths to reference assemblies.</param>
+        /// <param name="preprocessorSymbols">A set of preprocessor symbols to define.</param>
+        /// <param name="pluginPaths">The paths to plugins.</param>
+        /// <param name="intermediateOutputDirectory">The path to the directory that contains generated source files.</param>
+        public CompilationGenerator(string projectDirectory, IReadOnlyList<string> compile, IReadOnlyList<string> referencePath, IEnumerable<string> preprocessorSymbols, IReadOnlyList<string> pluginPaths, string intermediateOutputDirectory)
+        {
```
AArnott, line 45: This is technically a public entrypoint so I would add input validation to these arguments. 
However if the use case is that this is only public for calling between assemblies defined in this repo (i.e. the assembly isn't a ref assembly that we offer in a nuget package), and the caller also has NRT turned on, then we can omit them here.

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
AArnott: But the caller can't screw this up. This isn't argument validation but internal state validation, which the class now is (supposedly) guaranteeing is already correct. So I don't mind removal of this, *if* indeed the arg validation is added to the public constructor.

src/CodeGeneration.Roslyn.Tool/Program.cs
```diff
@@ -39,9 +40,9 @@ private static async Task<int> Core(string[] args, CancellationToken cancellatio
             IReadOnlyList<string> refs = Array.Empty<string>();
             IReadOnlyList<string> preprocessorSymbols = Array.Empty<string>();
             IReadOnlyList<string> plugins = Array.Empty<string>();
-            string generatedCompileItemFile = null;
-            string outputDirectory = null;
-            string projectDir = null;
+            string? generatedCompileItemFile = null;
+            string outputDirectory = null!;
+            string projectDir = null!;
```
AArnott, line 45: These two assignments are straight-up violations of NRT. No excuse for them, IMO. Please use `string?` here, or explain why it can't be done.

manne: Good catch. Based on the `require` parameter in `syntax.DefineOption`, I implicitly derived that the "command line" functionality will check for `null`.
I will add null checks.

manne: I added a comment to each assignment. And an explicit null check.
@AArnott Is this OK for you?

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -34,29 +33,48 @@ public class CompilationGenerator
         private readonly Dictionary<string, (PluginLoader loader, Assembly assembly)> cachedPlugins = new Dictionary<string, (PluginLoader, Assembly)>(StringComparer.OrdinalIgnoreCase);
```
manne, line 33:
```suggestion
        private readonly Dictionary<string, (PluginLoader Loader, Assembly Assembly)> cachedPlugins = new Dictionary<string, (PluginLoader, Assembly)>(StringComparer.OrdinalIgnoreCase);
```
@AArnott The style cop analyzer introduced a new rule [SA1316.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1316.md). Should I change according the rule?

AArnott: Let's go ahead and fix them. But let's set stylecop.json to allow for PascalCasing per [configuration](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#tuple-element-names).

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -74,22 +92,22 @@ public class CompilationGenerator
...
-        /// Gets or sets the directory with the project file.
+        /// Gets the directory with the project file.
         /// </summary>
-        public string ProjectDirectory { get; set; }
+        public string ProjectDirectory { get; }
 
         /// <summary>
         /// Runs the code generation as configured using this instance's properties.
         /// </summary>
         /// <param name="progress">Optional handler of diagnostics provided by code generator.</param>
```
AArnott, line 102: The doc says it's optional. But with your change it is no longer optional.

src/CodeGeneration.Roslyn.Tests.Generators/AddGeneratedExternGenerator.cs
```diff
@@ -26,7 +26,7 @@ public class AddGeneratedExternGenerator : RichBaseGenerator
         public AddGeneratedExternGenerator(AttributeData attributeData)
             : base(attributeData)
         {
-            Extern = (string)AttributeData.ConstructorArguments[0].Value;
+            Extern = (string)AttributeData.ConstructorArguments[0].Value!;
```
AArnott, line 29: I wonder if this non-null assumption is wise. Even if the attribute's own ctor checks for null, we're not accessing the value after it passes through the ctor -- we're accessing the argument before it's passed in. The compiler hasn't even run yet, so I don't think anything stops the user from passing null in and then we'd dereference that.
Should we check and throw instead?

src/CodeGeneration.Roslyn.Tests.Generators/MultiplySuffixGenerator.cs
```diff
@@ -37,7 +37,7 @@ public Task<SyntaxList<MemberDeclarationSyntax>> GenerateAsync(TransformationCon
                         var propertySymbol = context.SemanticModel.GetDeclaredSymbol(x);
                         var attribute = propertySymbol?.GetAttributes()
                             .FirstOrDefault(a => a.AttributeClass.Name == nameof(TestAttribute));
-                        string suffix = "Suff" + string.Concat(attribute?.NamedArguments.Select(a => a.Value.Value.ToString()) ?? Enumerable.Empty<string>());
+                        string suffix = "Suff" + string.Concat(attribute?.NamedArguments.Select(a => a.Value.Value!.ToString()) ?? Enumerable.Empty<string>());
```
AArnott, line 40: Similar here: I don't think assuming the user didn't pass in null here is safe. Should we check for null and throw a more helpful exception message instead?

AArnott: Ah, but this is in tests, so I don't care.

src/CodeGeneration.Roslyn.Tests.Generators/AddGeneratedExternGenerator.cs
AArnott: Never mind. This is just in tests, so I don't care.

src/CodeGeneration.Roslyn.Tool/Program.cs
```diff
@@ -39,9 +40,12 @@ private static async Task<int> Core(string[] args, CancellationToken cancellatio
...
             IReadOnlyList<string> plugins = Array.Empty<string>();
-            string generatedCompileItemFile = null;
-            string outputDirectory = null;
-            string projectDir = null;
+            string? generatedCompileItemFile = null;
+
+            // a value is assigned in the command line declaration (DefineOption), after option parsing there is an explicit null check
+            string outputDirectory = null!;
+            // a value is assigned in the command line declaration (DefineOption), after option parsing there is an explicit null check
+            string projectDir = null!;
```
AArnott, line 48: The explicit null check that comes later is good. But it doesn't mean we should define this as nullable up here, where we're deliberately setting it to null. If we set it to null, it should be nullable. Later, after the null check, the compiler will recognize that the value is not null and treat it as a non-null value, which should mean there's no benefit to making this `string` here instead of `string?`

src/CodeGeneration.Roslyn.Attributes/CodeGeneration.Roslyn.Attributes.csproj
```diff
@@ -6,6 +6,7 @@
     <Description>A package that offers attributes to drive code generation. API:
 * CodeGeneration.Roslyn.CodeGenerationAttributeAttribute
     </Description>
+    <Nullable>enable</Nullable>
```
amis92, line 9: Since it's added to Directory.Build.props, this is no longer necessary I think.

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
amis92: Seems to have been added in ctor.

src/CodeGeneration.Roslyn.Tool/Program.cs
```diff
@@ -72,15 +74,13 @@ private static async Task<int> Core(string[] args, CancellationToken cancellatio
...
             {
-                ProjectDirectory = projectDir,
-                Compile = Sanitize(compile),
-                ReferencePath = Sanitize(refs),
-                PreprocessorSymbols = preprocessorSymbols,
-                PluginPaths = Sanitize(plugins),
-                IntermediateOutputDirectory = outputDirectory,
-            };
+                Console.Error.WriteLine("The project directory must be specified.");
+                return 3;
```
amis92, line 80: Why is this changed to be required? ProjectDir is `string?` all the way down to TransformationContext.

manne: The commandline declaration states that `projectDir` is required. Therefore I implied, that in this case (from the generator's perspective) the project directory must not be `null`.

```CSharp
syntax.DefineOption("projectDir", ref projectDir, true, "The absolute path of the directory where the project file is located");
```
```

Comment (amis92) [reply src/CodeGeneration.Roslyn.Tool/Program.cs:80]: Interesting. Thanks for clarification.

File: src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs [hunk 12]
```diff
@@ -67,7 +69,12 @@ public CompilationGenerator(string? projectDirectory, IReadOnlyList<string> comp
         public IEnumerable<string> PreprocessorSymbols { get; }
 
         /// <summary>
-        /// Gets the paths to plugins.
+        /// Gets the build properties to expose to generators.
+        /// </summary>
+        public IReadOnlyDictionary<string, string> BuildProperties { get; }
+
+        /// <summary>
+        /// Gets or the path to the directory that contains generated source files.
```
amis92, line 77: Should be ` Gets the paths to plugins.`

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -31,37 +30,58 @@ public class CompilationGenerator
...
+        /// <param name="buildProperties">The build properties to expose to generators.</param>
+        public CompilationGenerator(string? projectDirectory, IReadOnlyList<string> compile, IReadOnlyList<string> referencePath, IEnumerable<string> preprocessorSymbols, IReadOnlyList<string> pluginPaths, string intermediateOutputDirectory, IReadOnlyDictionary<string, string> buildProperties)
+        {
+            ProjectDirectory = projectDirectory;
+            Compile = compile ?? throw new ArgumentNullException(nameof(compile));
+            ReferencePath = referencePath ?? throw new ArgumentNullException(nameof(referencePath));
+            PreprocessorSymbols = preprocessorSymbols ?? throw new ArgumentNullException(nameof(preprocessorSymbols));
+            PluginPaths = pluginPaths ?? throw new ArgumentNullException(nameof(pluginPaths));
+            IntermediateOutputDirectory = intermediateOutputDirectory ?? throw new ArgumentNullException(nameof(intermediateOutputDirectory));
+            BuildProperties = buildProperties;
```
amis92, line 53: I think a null check for consistency is in order :)

src/CodeGeneration.Roslyn.Tests/Helpers/CompilationTestsBase.cs
```diff
@@ -2,7 +2,9 @@
 // Licensed under the MS-PL license. See LICENSE.txt file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Collections.ObjectModel;
```
amis92, line 7: Where did that come from? :o

manne: reintegration of master

src/CodeGeneration.Roslyn.Engine/CompilationGenerator.cs
```diff
@@ -31,37 +30,58 @@ public class CompilationGenerator
...
         /// </summary>
-        public IReadOnlyDictionary<string, string> BuildProperties { get; set; }
+        public IReadOnlyDictionary<string, string> BuildProperties { get; }
 
         /// <summary>
-        /// Gets or sets the path to the directory that contains generated source files.
+        /// Gets the path to the directory that contains generated source files.
         /// </summary>
-        public string IntermediateOutputDirectory { get; set; }
+        public IReadOnlyList<string> PluginPaths { get; }
```
amis92, line 79: The comment still is wrong. It's a comment for `IntermediateOutputDirectory`. It should read
>  Gets the paths to plugins.

amis92: Fixed myself.

PR #228, opened 2020-04-28
feat: Allow override of the generated files output path

### LINKED ISSUE #95: Generate files in project instead of obj directory
In many cases it's desirable to save generated code in repository. It's a common practice for binaries, see recomendation for different `lock` files in different ecosystems.

It seems reasonable to have some setting like `bool GenerateFilesInProject` that could enable in-project generation, similar to how WinForm does. It would be nice to have access to generated files in repository, without rebuilding the whole solution.

I can make it myself if we are agreed on that, so my changes wouldn't be futile.

robkroll:
Allow override of the output directory where CGR.Tool writes generated files. Defaults to existing value `$(IntermediateOutputPath)` so no change for existing users.

Partially fixes #95

### DIFF
--- a/src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
+++ b/src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
@@ -64,6 +64,8 @@
       <CodeGenerationRoslynToolPath
         Condition="'$(GenerateCodeFromAttributesToolPathOverride)' != ''">$(GenerateCodeFromAttributesToolPathOverride)</CodeGenerationRoslynToolPath>
       <CodeGenerationRoslynToolPath>$(CodeGenerationRoslynToolPath.Trim())</CodeGenerationRoslynToolPath>
+      <CodeGenerationRoslynFilesOutputPath
+        Condition="'$(CodeGenerationRoslynFilesOutputPath)' == ''">$(IntermediateOutputPath)</CodeGenerationRoslynFilesOutputPath>
       <_CodeGenToolOutputBasePath>$(IntermediateOutputPath)$(MSBuildProjectFile).dotnet-codegen</_CodeGenToolOutputBasePath>
       <_CodeGenToolResponseFileFullPath>$(_CodeGenToolOutputBasePath).rsp</_CodeGenToolResponseFileFullPath>
       <_CodeGenToolGeneratedFileListFullPath>$(_CodeGenToolOutputBasePath).GeneratedFileList.txt</_CodeGenToolGeneratedFileListFullPath>
@@ -73,7 +75,7 @@
         @(_CodeGenerationRoslynResolvedProperty->'%(Key)=%(Value)'->Replace(';', '%3B')->'--buildProperty;%(Identity)');
         @(CodeGenerationRoslynPlugin->'--plugin;%(Identity)');
         --out;
-        $(IntermediateOutputPath);
+        $(CodeGenerationRoslynFilesOutputPath);
         --projectDir;
         $(MSBuildProjectDirectory);
         --generatedFilesList;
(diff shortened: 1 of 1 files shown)


src/CodeGeneration.Roslyn.Tool/build/CodeGeneration.Roslyn.Tool.targets
```diff
@@ -64,6 +64,8 @@
       <CodeGenerationRoslynToolPath
         Condition="'$(GenerateCodeFromAttributesToolPathOverride)' != ''">$(GenerateCodeFromAttributesToolPathOverride)</CodeGenerationRoslynToolPath>
       <CodeGenerationRoslynToolPath>$(CodeGenerationRoslynToolPath.Trim())</CodeGenerationRoslynToolPath>
+      <CodeGenerationRoslynFilesOutputPath
```
amis92, line 67: I think I'd prefer `CodeGenerationRoslynToolOutputPath`.

amis92:
I'd also like to suggest that maybe a warning should be raised (with a code to help suppression) when:
1. There are multiple TargetFrameworks specified and TargetFramework value is not part of the specified Path.
2. `Configuration` value is not part of the Path.

These two cases will result in files being overwritten and an undefined result in the end.

robkroll:
Thanks for the feedback, I'll work on those changes. One question in the meantime: why does CGR generate an AssemblyAttributes file (.e.g _.NETStandard,Version=v2.0.AssemblyAttributes.APNPx2rD.generated.cs_). 

The file content is only the auto-generated banner and then:
```
using System;
using System.Reflection;
```

But it seems that this file is picked up for generation again by CGR. So if the `CodeGenerationRoslynToolOutputPath` folder is set to `$(MsBuildProjectFolder)` it will regenerate this file over and over (appending .generated.generated...) until it reaches the 260 char limit and breaks.

amis92:
> So if the CodeGenerationRoslynToolOutputPath folder is set to $(MsBuildProjectFolder) it will regenerate this file over and over (appending .generated.generated...) until it reaches the 260 char limit and breaks.

This will be the case anyway if you'll not exclude those files from default Compile include manually. This should also probably be highlighted in documentation.

CGR runs before CoreCompile (at which time the AssemblyAttributes.cs is already generated), and it takes in all Compile items (`ItemGroup->Compile`).

So, by default, if you set the `CodeGenerationRoslynToolOutputPath` to `$(MsBuildProjectFolder)`, only running build once is a supported scenario (and only if you're no multitargeting).

Because every subsequent build will include the generated files in the Compile item by default, and we'll generate duplicates of those files, as you've noticed.

Our compilation setup wasn't ever meant to have the files saved to the project directory. It's designed from the ground up as a per-Compilation runner.

robkroll:
>CGR runs before CoreCompile (at which time the AssemblyAttributes.cs is already generated), and it takes in all Compile items (ItemGroup->Compile).

Ok, so all files are passed to CGR.Tool before the compile. But if a generator outputs, say, an empty class definition (and the generated class is therefore _not_ attributed for code generation) it will get included in a subsequent call to CGR.Tool, but it will skip the file generation (or overwrite it) because the file name including hash is the same.

Only if a generated file is attributed for code generation do we get this recursive _.generated.generated.generated..._ problem. For example this happens with the `DuplicateWithSuffixGenerator` from the example. 

Except for the `AssemblyAttributes.cs` file, this seems to be generated regardless.

amis92:
> as it's not marked for generation it's not passed to a generator and nothing happens with it -- no new files are produced

Speculating here (no proof), but I'm pretty sure we create the new files no matter if there were any generators run or trigger attributes found. It may be the case that they get re-generated and overwritten. It all just adds to this being a totally not supported scenario. :(

amis92:
I think we could mitigate at least partially the infinite-regeneration scenario by changing slightly how we consume the Compile list:
1. Map input paths to generated paths
2. Exclude input paths that exist in the generated paths collection.
3. Generate g.cs only for items left on Compile (filtered).

@robkroll would you be interested in adding this in this PR?

@AArnott do you think this'd work?

AArnott:
Yes, I think that's promising @amis92. Another option, IIRC, is that we only generate for "all" source files right now because of an `ItemDefinitionGroup` that turns it on by default for all `Compile` items. A project that's doing something special like @robkroll  proposes likely wants to turn this default off and explicitly select with metadata which Compile items should actually result in code generation.

Issue #229, opened 2020-05-06
How does this relate to Source Generators from MS?


ugumba:
https://devblogs.microsoft.com/dotnet/introducing-c-source-generators

Will CodeGeneration.Roslyn be made redundant, or are there any important differences?

amis92:
Source Generators will make this project redundant indeed.

There are differences, but not considerable enough to give this project any life support in future.

garyng:
Why this makes me feel sad?

AArnott:
I'm glad this project has been helpful to you, @garyng.

LokiMidgard:
@AArnott  as you said [you don't want to maintain this anymor](https://github.com/AArnott/CodeGeneration.Roslyn/issues/237#issuecomment-741485572), maybe mention that in the README. I think you can also set something in github to mark it as no longer maintained.

AArnott:
@amis92 Are you ready to archive this project as well? Or are you interested in maintaining it any longer?

amis92:
I am ready to archive it.

![https://images.app.goo.gl/BpvYeLQGtcmFPRrp6]

Let's throw in a couple of links to the Source Generators into the Readme and call it done ☺️

AArnott:
Let the record show that @amis92 really carried this project forward for the last 2+ years. Mega thanks!

PR #232, opened 2020-06-16
Provide distinct Usings for final generated compliation unit

stevenknox:
When using multiple different generators that have an overlap of Using Directives , the resulting consumer project may get a CS0105 warning "The using directive for 'Foo' appeared previously in this namespace"

I have applied Distinct() to ensure there are no duplicates in the final generated compilation.

### DIFF
--- a/src/CodeGeneration.Roslyn.Engine/DocumentTransform.cs
+++ b/src/CodeGeneration.Roslyn.Engine/DocumentTransform.cs
@@ -110,7 +110,7 @@ namespace CodeGeneration.Roslyn.Engine
             var compilationUnit =
                 SyntaxFactory.CompilationUnit(
                         SyntaxFactory.List(emittedExterns),
-                        SyntaxFactory.List(emittedUsings),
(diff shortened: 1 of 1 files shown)


AArnott (commented):
Often in Roslyn's OM, the types don't override the Equals method with by-value semantics. Have you confirmed that this change is effective across multiple generators mentioned in your scenario? If so, it looks good.

amis92:
Please add a test based on 

https://github.com/AArnott/CodeGeneration.Roslyn/blob/fd1200769415b5428fbd6c5859dbe55cf78f0b7b/src/CodeGeneration.Roslyn.Tests.Generators/AddGeneratedUsingGenerator.cs

and

https://github.com/AArnott/CodeGeneration.Roslyn/blob/master/src/CodeGeneration.Roslyn.Tests/DocumentTransformTests.cs

Shouldn't be too complicated :)

stevenknox:
Perfect! ill get a test added for this first thing in the morning =)

AArnott:
This project is getting archived in favor of Source Generators. See #229

